PR #4, opened 2016-08-30
Chainable API

### LINKED ISSUE #2: How to use
I am wondering what is the best way to use `boot-in-the-arse`. Currently there are two options:

a. subclassing
b. passing another server through the constructor

Case a. works as expected, but it fills the object with the `boot-in-the-arse` properties. It would probably need to be more documented.

Case b. uses the passed object as the `context` and `this` in the plugin loading. However it does not expose `use`, the `'start'` event and the upcoming `ready()`  and `after()` #1.

I think there should be a _blessed_ way on using this.
What should we do?

I proposed we auto-bind `use`, `ready` and `after` in case b), so that they work as expected.

cc @mcdonnelldean @lucamaraschi

mcollina:
@mcdonnelldean @lucamaraschi please review this. It should provide what we need.

Fixes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,23 +49,40 @@ boot.on('start', function () {
 
 ### boot([instance], [started])
 
-Start a booting sequence. The `boot` function can be used also as a
-constructor to inherits from.
+Start a booting sequence.
 
 `instance` will be used as the first
-argument of all plugins loaded, you are responsible for exposing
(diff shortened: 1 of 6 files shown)


lucamaraschi:
👍  LGTM

I like the easy and intuitive interface.

PR #9, opened 2017-02-24
Added app.override() to create plugin inheritance chain

mcollina:
cc @delvedor

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ function third (instance, opts, cb) {
   * <a href="#use"><code>instance.<b>use()</b></code></a>
   * <a href="#after"><code>instance.<b>after()</b></code></a>
   * <a href="#ready"><code>instance.<b>ready()</b></code></a>
+  * <a href="#override"><code>instance.<b>override()</b></code></a>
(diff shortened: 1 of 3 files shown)


delvedor:
LGTM, if you want you can add the following test case for override, which is specific for the [fastify](https://github.com/fastify/fastify) use case.

```js
test('fastify test case', (t) => {
  t.plan(7)

  const noop = () => {}

  function build () {
    const app = boot(server, {})
    app.override = function (s) {
      return Object.create(s)
    }

    server.add = function (name, fn, cb) {
      if (this[name]) return cb(new Error('already existent'))
      this[name] = fn
      cb()
    }

    return server

    function server (req, res) {}
  }

  const instance = build()
  t.ok(instance.add)
  t.ok(instance.use)

  instance.use((i, opts, cb) => {
    t.notEqual(i, instance)
    t.ok(instance.isPrototypeOf(i))

    i.add('test', noop, (err) => {
      t.error(err)
      t.ok(i.test)
      cb()
    })
  })

  instance.ready(() => {
    t.notOk(instance.test)
  })
})
```

PR #10, opened 2017-03-18
Override

delvedor:
Added `'override'` option to force the ovverride of the instance.


delvedor:
Related to #9.

delvedor:
After a discussion, we found that the solution proposed in #9 is enough.

Issue #6, opened 2016-10-03
Pass context to ready and after


mcdonnelldean:
This allows these functions to be written in other modules. consider the following.

``` js
// some setup function that handles post plugin
// loading for some icky plugin.
module.exports = function (context, done) {
  // I can use the context without issue now
  context.dispatch('comms:open')
  done()
}

```

mcollina:
Yeah! PR?

mcdonnelldean:
It's on the list :D

mcollina:
Fixed in v0.5.0.

PR #12, opened 2017-03-20
Updated override and after/ready callbacks

delvedor:
- `override` takes as second parameter the plugin function
- if `after`/`ready` have two parameters, the first one is the context (this to avoid breaking changes) #6

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -71,6 +71,9 @@ function Boot (server, opts, done) {
     this.once('start', done)
   }
 
+  /* function readyQIterator (func, cb) {
+    callWithCbOrNextTick(func, cb, server)
+  } */
   this._readyQ = fastq(this, callWithCbOrNextTick, 1)
   this._readyQ.pause()
   this._readyQ.drain = () => {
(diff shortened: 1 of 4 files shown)


boot.js
```diff
@@ -187,7 +188,7 @@ function Plugin (parent, func, opts, callback) {
 
 Plugin.prototype.exec = function (server, cb) {
   const func = this.func
-  this.server = this.skipOverride ? server : this.parent.override(server)
+  this.server = this.skipOverride ? server : this.parent.override(server, func)
```
mcollina: can we just remove `skipOverride`, and use it in fastify?

boot.js
```diff
@@ -217,12 +218,18 @@ function loadPlugin (toLoad, cb) {
...
+  }
+
   if (func.length === 0) {
     func()
     process.nextTick(cb)
-  } else {
+  } else if (func.length === 1) {
     func(cb)
+  } else {
+    func(context, cb)
```
mcollina: why this change?

delvedor: I was trying to implement #6 without introducing a breaking change.

boot.js
delvedor: Yeah, maybe we should :)

PR #13, opened 2017-03-30
After/Ready error handling

delvedor:
Implementation of the discussion done in https://github.com/fastify/fastify/issues/59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,23 +148,24 @@ The functions will be loaded in the same order as they are inside the array.
 -------------------------------------------------------
 <a name="after"></a>
 
-### app.after(func([context], [done]), [cb])
+### app.after(func(error, [context], [done]), [cb])
 
 Calls a function after all the previously defined plugins are loaded, including
 all their dependencies. The `'start'` event is not emitted yet.  
-If one parameter is given to the callback, that parameter will be the `done` callback.  
-If two parameters are given to the callback, the first will be the `contenxt`, the second will be the `done` callback.
+1. If one parameter is given to the callback, that parameter will be the `error` object.  
+2. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.  
+3. If three parameters are given to the callback, the first will be the `error` object, the second will be the `context` and the third the `done` callback.
 
 ```js
 const server = {}
 ...
 // after with one parameter
-boot.after(function (done) {
+boot.after(function (err, done) {
   done()
 })
 
 // after with two parameters
-boot.after(function (context, done) {
+boot.after(function (err, context, done) {
   assert.equal(context, server)
   done()
 })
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -148,23 +148,24 @@ The functions will be loaded in the same order as they are inside the array.
...
+1. If one parameter is given to the callback, that parameter will be the `error` object.  
+2. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.  
+3. If three parameters are given to the callback, the first will be the `error` object, the second will be the `context` and the third the `done` callback.
 
 ```js
 const server = {}
 ...
 // after with one parameter
-boot.after(function (done) {
+boot.after(function (err, done) {
```
mcollina: the comment needs to be updated

README.md
```diff
@@ -148,23 +148,24 @@ The functions will be loaded in the same order as they are inside the array.
...
   done()
 })
 
 // after with two parameters
```
mcollina: the comment needs to be updated

README.md
```diff
@@ -148,23 +148,24 @@ The functions will be loaded in the same order as they are inside the array.
 -------------------------------------------------------
 <a name="after"></a>
 
-### app.after(func([context], [done]), [cb])
+### app.after(func(error, [context], [done]), [cb])
```
mcollina: I do not think `after()`  should be called with an error at all. We should bail off the plugin loading.

delvedor: Why not, in this way the user can handle the error inside the after.
Example:
If plugin 1 fails => load plugin 2.

mcollina: that should be in the `register` callback, not in `after` IMHO.

delvedor: Agree, but at the moment this is the behaviour:
```js
test('error', (t) => {
  t.plan(2)

  const server = { my: 'server' }
  const app = boot(server)

  app.use(function (s, opts, done) {
    done(new Error('err'))
  }, err => {
    t.ok(err)
    app.use(function (s, opts, done) {
      done()
    })
  })

  app.ready(function (err) {
    t.ok(err)
  })
})
```
As you can see, we have not that capability, because the error will arrive to the ready callback in every case.
With after I can "block" that error ina  specific point, as you can see here:
```js
test('error', (t) => {
  t.plan(3)

  const server = { my: 'server' }
  const app = boot(server)

  app.use(function (s, opts, done) {
    done(new Error('err'))
  }, err => {
    t.ok(err)
  })

  app.after(function (err) {
    t.ok(err)
    app.use(function (s, opts, done) {
      done()
    })
  })

  app.ready(function (err) {
    t.error(err)
  })
})
```
This is why yesterday I was saying that we should raise the error in the ready callback only if the error is not handled in the use callback.

mcollina: OK for this behavior you are describing right know.

delvedor: Sorry, I didn't understand, do you want me to implement this behaviour?
> This is why yesterday I was saying that we should raise the error in the ready callback only if the error is not handled in the use callback.

mcollina: > This is why yesterday I was saying that we should raise the error in the ready callback only if the error is not handled in the use callback.

yes.

PR #14, opened 2017-03-31
Docs

delvedor:
Some updates to the docs.
I also added myself as collaborator inside the package.json.

Do you think we can add something more?
For example I think we can add a part that explain better how the *lifecycle* of avvio works.


mcollina:
I think it's ok :)

PR #16, opened 2017-07-10
Close queue

delvedor:
With this PR we introduce a new queue to handle the shutdown procedure, following the *avvio* style!

- Use `onClose` to add new functions to execute.
- Use `close` to start the process

Related: https://github.com/fastify/fastify/issues/123

Feedbacks?

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -171,7 +171,7 @@ Boot.prototype.after = function (func, cb) {
 }
 
 Boot.prototype.onClose = function (func) {
-  this._closeQ.push(func, err => {
+  this._closeQ.unshift(func, err => {
     if (err) this._error = err
   })
   return this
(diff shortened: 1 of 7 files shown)


boot.js
```diff
@@ -164,68 +170,25 @@ Boot.prototype.after = function (func, cb) {
   return this
 }
 
-Boot.prototype.ready = function (func) {
-  this._readyQ.push(func)
+Boot.prototype.onClose = function (func) {
+  this._closeQ.push(func, err => {
+    if (err) this._error = err
+  })
```
mcollina: I think we should call [`unshift()`](https://www.npmjs.com/package/fastq#unshift) here. More or less, the latest you register a shutdown part, the sooner it will be executed. This is the inverse of the start phase.

delvedor:
> LGTM as semver-major.

Why semver major?
I don't see a breaking change here.

mcollina:
adding the new properties to the instance is.

delvedor:
Ok I got you, it is not a major change for avvio itself, but it is for the apps that are using it 👍

PR #18, opened 2017-07-13
Do not erase error if callback is called sync

### LINKED ISSUE #17: Error are not passed from after to ready
I don't think this should happen.
Could be related to [this](https://github.com/mcollina/avvio/blob/master/boot.js#L177)?

Example:
```js
test('shold pass the errors from after to ready', (t) => {
  t.plan(6)

  const server = {}
  const app = boot(server, {})

  server.use(function (s, opts, done) {
    t.equal(s, server, 'the first argument is the server')
    t.deepEqual(opts, {}, 'no options')
    done()
  }).after((err, done) => {
    t.error(err)
    done(new Error('some error'))
  })

  server.onClose(() => {
    t.ok('onClose called')
  })

  server.ready(err => {
    t.is(err.message, 'some error') // this will fail
  })

  app.on('start', () => {
    server.close(() => {
      t.pass('booted')
    })
  })
})
```

mcollina:
Fixes #17

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -174,8 +174,6 @@ Boot.prototype._addPlugin = function (plugin, opts, callback) {
 }
 
 Boot.prototype.after = function (func, cb) {
-  // TODO do not rely on .use()
-  // const server = this._server
   this.use(function (s, opts, done) {
     callWithCbOrNextTick.call(this, func, done)
   }.bind(this), cb)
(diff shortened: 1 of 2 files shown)


boot.js
```diff
@@ -207,16 +205,19 @@ function callWithCbOrNextTick (func, cb, context) {
     context = this._server
   }
 
+  var err = this._error
```
delvedor: I think you should wrap this inside an `if (this)`, as was done few lines above :)
Or, if not needed, remove the `if (this)` few lines above :P

mcollina: done

PR #19, opened 2017-07-13
Expose the error only inside the close callback

delvedor:
As titled


delvedor:
Yes, I've also added a test to cover that.

mcollina:
LGTM

PR #22, opened 2017-09-18
Updated onClose signature handling

delvedor:
As titled, in this way the we conform the `onClose` api to `ready`, `close` etc.
I think this should be considered a semver major.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ Calls a function after all the plugins and `after` call are completed, but befor
 The callback changes basing on the parameters your are giving:
 1. If one parameter is given to the callback, that parameter will be the `error` object.  
 2. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.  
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -324,14 +324,20 @@ app.use(function first (s1, opts, cb) {
 Registers a new callback that will be fired once then `close` api is called.
 
 The callback changes basing on the parameters your are giving:
-1. If one parameter is given to the callback, that parameter will be the `context`.  
-2. If two parameters are given to the callback, the first will be the `context`, the second will be the `done` callback.  
+1. If no parameters are provided, the function will be considered synchronous.
+2. If one parameter is given to the callback, that parameter will be the `done` callback.  
+3. If two parameters are given to the callback, the first will be the `context`, the second will be the `done` callback.
```
mcollina: we need to specify here and in `ready()`, that it is not the current context, but only the top-level one.

PR #23, opened 2017-09-18
Updated encapsulation handling

delvedor:
We are not passing the correct encapsulated instance when using `override`, this pr aims to fix that.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -30,23 +30,23 @@ function wrap (server, opts, instance) {
     return this
   }
 
-  server[afterKey] = function (cb) {
-    instance.after(cb)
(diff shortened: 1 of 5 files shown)


boot.js
```diff
@@ -257,21 +259,7 @@ function encapsulateTwoParamServer (func, that) {
...
-    } else if (func.length === 2) {
-      func(err, cb)
-    } else {
-      func(err, this._server, cb)
-    }
-  }
-}
-
-function encapsulateThreeParamServer (func, that, useContext) {
+function encapsulateThreeParam (func, that, useContext) {
```
mcollina: `useContext` is not "used" here anymore

PR #28, opened 2017-09-29
Async await support within use

mcollina:
As titled

cc @allevo


allevo:
I'd like to build a npm module to handle this automatically. Maybe this weekend I figure out how to do this. In this case, are you Interested in that npm?

mcollina:
this seems pretty custom, but I’ll be happy to see what the ecosystem offers.

delvedor:
IMHO what we should make very clear, is that `use` must be called synchronously.
For example the following should not be valid:
```js
getStruff()
  .then(stuff => {
    app.use(stuff)
  })
```

We should be able to detect it and throw an error.
Or find a way to support it, but I don't like it so much, `avvio` should be used as single system to handle asynchronous bootstrap, we should discourage mixed approaches.

The above example for instance, should be like this:
```js
app.use(async function (app, opts, next) {
  const stuff = await getStruff()
  app.use(stuff, next)
})
```

mcollina:
@delvedor can you please provide another (bigger) example? It seems that is working.

allevo:
@mcollina try https://github.com/allevo/three-corners

PR #33, opened 2017-11-09
close must receive a function or a falsey value

mcollina:
If calling `.close()` with a non-function, throw.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -32,21 +32,33 @@ function wrap (server, opts, instance) {
   }
 
   server[afterKey] = function (func) {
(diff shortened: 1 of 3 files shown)


boot.js
```diff
@@ -207,6 +219,9 @@ Boot.prototype.onClose = function (func) {
 Boot.prototype.close = function (func) {
   this._error = null
   if (func) {
+    if (func && typeof func !== 'function') {
```
delvedor: Here just the `typeof` is enough.

Issue #24, opened 2017-09-18
Improve documentation


delvedor:
At the moment the documentation is almost complex as the actual code, we should fix that.

HelloEdit:
Also for the `app.use(func, [opts])`:
Loads one or more functions asynchronously. **The function must have the signature: instance, options, done**

Is it an inconsistency or misunderstanding on my part that the asynchronous function does not respect the signature?

```js 
async function plugin (server, opts) {} // without done
```

minemax-ua:
The documentation definitely needs a complete overhaul.

mcollina:
Would you like to help?

minemax-ua:
> Would you like to help?

I prefer to help with something I know very well, so it's rather No.

Tony133:
Resolved in PR #362. If there's anything else that could be improved, feel free to open a new issue.

PR #34, opened 2017-11-10
Replace legacy `boot` variable with the `app` avvioo instance

marcbachmann:
The `boot` variable was still in use since the module got renamed and caused confusion for some people. This change makes the documentation a bit more readable.

There's still https://github.com/mcollina/avvio/issues/24 to tackle.


marcbachmann:
FYI We'll soon have another loader with a different approach we currently used at @upfrontIO.
@peyerluk or myself will work on that.

mcollina:
Fantastic, thanks!

PR #35, opened 2017-11-13
After should passthrough the errors

delvedor:
As titled.
Related: https://github.com/fastify/fastify/pull/454 https://github.com/fastify/fastify/issues/460


delvedor:
I think we should consider this as a minor/bugfix.
Take https://github.com/fastify/fastify/pull/454 for example, the issue were that we was not handling the error, and we failed silently.

PR #36, opened 2017-11-25
Added `wait` api

delvedor:
As titled, in this way we can provide a better support for *async-await*.
See https://github.com/fastify/fastify/issues/481 for context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ async function third (instance, opts) {
   * <a href="#use"><code>instance.<b>use()</b></code></a>
   * <a href="#after"><code>instance.<b>after()</b></code></a>
   * <a href="#ready"><code>instance.<b>ready()</b></code></a>
+  * <a href="#wait"><code>instance.<b>wait()</b></code></a>
   * <a href="#override"><code>instance.<b>override()</b></code></a>
   * <a href="#onClose"><code>instance.<b>onClose()</b></code></a>
   * <a href="#close"><code>instance.<b>close()</b></code></a>
@@ -262,9 +263,22 @@ app.ready(function (err, context, done) {
 Returns the instance on which `ready` is called, to support a chainable API.
 
 -------------------------------------------------------
+<a name="wait"></a>
+
+### app.wait()
+If you are using async await, you can avoid the use of `after` and `ready`, and just use the `wait` api which returns a promise.
+```js
+
+await app.use(plugin).wait()
+
+async function plugin (instance, opts) {
+  console.log('plugin')  
+}
+```
+-------------------------------------------------------
 <a name="express"></a>
 
-### avioo.express(app)
+### avvio.express(app)
 
 Same as:
 
--- a/boot.js
+++ b/boot.js
@@ -206,6 +206,23 @@ Boot.prototype.after = function (func, cb) {
   return this
 }
 
+Boot.prototype.wait = function () {
+  return new Promise((resolve, reject) => {
+    this.use(_wait.bind(this))
+
+    function _wait (s, opts, done) {
+      process.nextTick(done)
+      var err = this._error
+      this._error = null
+      if (err) {
+        reject(err)
+      } else {
+        resolve()
+      }
+    }
+  })
+}
+
 Boot.prototype.onClose = function (func) {
   this._closeQ.unshift(func, callback.bind(this))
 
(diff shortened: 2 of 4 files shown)


StarpTech:
Nice :ok_hand:

test/basic.test.js
```diff
@@ -205,3 +205,47 @@ test('promise long resolve', (t) => {
     t.notOk(err)
   })
 })
+
+test('wait plugin registration', (t) => {
```
StarpTech: What about test to cover error propagations with async await ?

```js
app.use(async () => {})
```

delvedor: Take a look [here](https://github.com/mcollina/avvio/blob/74ac2125913658002a21169bae690f417e0cfe98/test/async-await.js#L145-L158) :D

StarpTech: I didn't saw it :D

boot.js
```diff
@@ -206,6 +206,23 @@ Boot.prototype.after = function (func, cb) {
   return this
 }
 
+Boot.prototype.wait = function (func) {
```
StarpTech: `func` can be removed

README.md
```diff
@@ -261,10 +262,23 @@ app.ready(function (err, context, done) {
 
 Returns the instance on which `ready` is called, to support a chainable API.
 
+-------------------------------------------------------
+<a name="wait"></a>
+
+### app.wait()
+If you are using async await, you can avoid the use of `after` and `ready`, and just use the `wait` api.
```
StarpTech: If you are using async await, you can avoid the use of `after` and `ready`, and just use the `wait` api which returns a promise.

StarpTech:
LGTM

mcollina:
Closing for now, feel free to reopen if you still want to pursue this idea.

allevo:
Can we reopen this discussion?

From developer point of view, this feature is very useful. For example:

```
module.exports = async function (fastify, options) {
  fastify.register(fastifyEnv, { ... })
  await fastify.wait()
  fastify.register(fastifyMongodb, { url: fastify.config.MONGODB_URL })
}
```

currently writable as:

```
module.exports = function (fastify, options, done) {
  fastify.register(fastifyEnv, { ... })
    .after(() => {
      fastify.register(fastifyMongodb, { url: fastify.config.MONGODB_URL })
        .after((err, done) => {
          if (err) return done(err)
          fastify.mongo.db.createCollection('myCollection', done)
        })
    })
  done()
}
```

delvedor:
The only way to support what you are proposing is to play a lot with the internal queues, which is not that easy and it will require a crazy amount of tests.

allevo:
@delvedor I know.

But the process start up is not possible without doing a callback hell...

mcollina:
How about?

```js
module.exports = async function (fastify, options) {
  fastify
     .register(fastifyEnv, { ... })
     .register(fastifyMongodb, { url: fastify.config.MONGODB_URL })
     .register(myRoutes)
}

async function myRoutes(f, opts) {
  fastify.mongo.db.createCollection('myCollection', done)
}
```

mcollina:
My bad, that's not possible.

mcollina:
or better:

```js
fastify
  .register(fastifyEnv)
  .register(addMongo)
  .register(myRoutes)

async function addMongo (f, opts) {
     f.register(fastifyMongodb, { url: fastify.config.MONGODB_URL })
}

async function myRoutes(f, opts) {
  fastify.mongo.db.createCollection('myCollection', done)
}
```

allevo:
This example doesn't work
```

async function decorateCollectionInstance (fastify) {
  fastify.decorate('collection', new MyAwesomeBusinessLogicClass(fastify.mongo.db)
  // TypeError: Cannot read property 'db' of undefined
}

async function mongodb (fastify) {
  fastify.register(fastifyMongodb, { url: fastify.config.MONGODB_URL })
  fastify.register(decorateCollectionInstance)
}

const envSchema = {
  type: 'object',
  required: ['MONGODB_URL'],
  properties: {
    MONGODB_URL: { type: 'string' }
  }
}
module.exports = async function (fastify, options) {
  fastify.register(fastifyEnv, { schema: envSchema, data: options, env: false })
    .register(mongodb)
    .register(decorateCollectionInstance)
}

```

mcollina:
@allevo can you send a PR for fastify with that example? That _should_ work.

PR #37, opened 2017-12-06
Add support for `ready` to return a promise.

coreyfarrell:


README.md
```diff
@@ -234,9 +234,11 @@ The callback changes basing on the parameters your are giving:
 3. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.
 4. If three parameters are given to the callback, the first will be the `error` object, the second will be the top level `context` unless you have specified both server and override, in that case the `context` will be what the override returns, and the third the `done` callback.
 
+If no callback is provided `ready` will return a Promise that is resolved or rejected once plugins and `after` calls are completed.  No data is provided to the `.then` callback, if an error occurs it is provided to the `.catch` callback.
```
mcollina: I think we should provide the context to `resolve`.

allevo:
Sorry for the intrusion

I'm 👍  for moving ready for a better async support

I'm 👎  for return different kind of type depending on a parameter presence

mcollina:
@allevo you propose to return `null` if we are not generating a `Promise`?

allevo:
> That'll be a semver-major change.

Yeah. And it could be a problem. But returning different type is very a bad thing. `null` could be a solution.

For example the `mongodb` library return a `Promise` or `undefined` (no return statement). See [here](http://mongodb.github.io/node-mongodb-native/2.2/api/lib_cursor.js.html#line831) for example

README.md
coreyfarrell: I've updated my working copy code to do this.  I'll update the docs and push the update once it's decided what should be returned from `app.ready(callbackForm)`.

mcollina:
If we do a semver-major, we might want to do some more work here.

coreyfarrell:
Should I make `ready` return `null` or `undefined` when not returning a `Promise`?

@delvedor So I make sure I understand correctly you want me to add a check to `Boot.prototype.use`:
```js
if (this.booted) {
  throw new Error('plugin cannot be loaded after ready')
}
```
Should that be done in a separate commit so it appears in `git log --oneline`?

mcollina:
> Should I make ready return null or undefined when not returning a Promise?

Yes, I'd prefer undefined.

@coreyfarrell the check that @delvedor wants is already part of https://github.com/mcollina/avvio/pull/38/files#diff-fe60a8aed4b3103661f9d88361346b0dR185. They will be released together anyway. Go ahead and just do the change on `ready()`.

coreyfarrell:
Returning undefined helped me catch the fact that I missed the `wrap()` function while writing this patch. I've fixed that for `ready` but it also shows that `tests/async-await.js` fails to test the wrap feature using:
```js
const app = {}
boot(app)
```

Right now I'm manually switching the tests in so I can find/fix the issues with `wrap` async.  I haven't looked into what it would take to have async-await.js run each test twice (once with and once without wrap).

coreyfarrell:
I'm now realizing github probably didn't send any notifications when I updated the PR yesterday.  I've addressed feedback and fixed issues found by testing with wrapped objects.

README.md
mcollina: This has not been updated

boot.js
```diff
@@ -229,8 +228,23 @@ Boot.prototype.close = function (func) {
 }
 
 Boot.prototype.ready = function (func) {
-  this._readyQ.push(func)
-  return this
+  if (func) {
```
mcollina: I would check if this is a function and throw if it exists but it is not.

test/chainable.test.js
```diff
@@ -16,10 +16,8 @@ test('chainable standalone', (t) => {
       done()
     }).ready(function () {
       t.pass('we are ready')
-    }).use(function (ctx, opts, done) {
-      t.pass('3rd plugin')
-      done()
```
mcollina: Can you restore this?

test/chainable.test.js
```diff
@@ -38,10 +36,8 @@ test('chainable automatically binded', (t) => {
...
```
mcollina: can you restore this?

test/chainable.test.js
coreyfarrell: `ready` no longer returns `this` so the test now verifies that it is not chainable.

If you are concerned about testing that `use` after `ready` does the right thing I believe other tests do this, `ready adds at the end of the queue` from `after-and-ready.test.js` for example calls ready before calling use.

Please let me know if I'm misunderstanding your finding.

PR #41, opened 2018-01-08
Removed callback from use

delvedor:
As titled.
This change will need a new major release.
Related: https://github.com/fastify/fastify/issues/460


mcollina:
Should we land this?

delvedor:
I've tested this with the master branch of Fastify and there are not failing test.
The current documentation already show that you can't use a callback within use.
Should we release this as major or not?

mcollina:
Integers are cheap.

PR #42, opened 2018-01-09
Remove use(array) form

mcollina:
cc @jsumners @starptech @allevo


allevo:
Can we throw an error if an array is supplied?

delvedor:
Please update also the docs.

mcollina:
@allevo We are already throwing.

I plan to release this as 5.0.1.

PR #44, opened 2018-01-17
After event should be fired before calling ready handler

allevo:
This issue was suggested in this comment https://github.com/fastify/fastify/pull/671#discussion_r162013010

I don't know if the current behavior is correct or not.
This PR adds a test for checking what I expected.

**NB:** I didn't found any tests that check that call order.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -278,3 +278,16 @@ test('throws correctly if registering after ready', (t) => {
     }, 'root plugin has already booted')
   })
 })
+
+test('`start` event should launch before the ready callback', (t) => {
+  t.plan(1)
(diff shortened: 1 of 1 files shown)


mcollina:
I fear this isn’t easy. I moved it in the latest refactoring to handle some edge case (which I forgot :/).

allevo:
@mcollina we are still in time for revert the commit for https://github.com/fastify/fastify/pull/671

mcollina:
I will leave that with `listening`, it's ok.

mcollina:
Is this still an issue? Should it be fixed?
I think this is not needed anymore, but let us know.

PR #45, opened 2018-01-28
Stop chain on failure

allevo:
As a developer, I want that `avvio` should stop the process start up if some plugin fails

This PR provides a simple test to start a discussion about this feature.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -278,3 +278,18 @@ test('throws correctly if registering after ready', (t) => {
     }, 'root plugin has already booted')
   })
 })
+
+test('failure use should stop the chain', (t) => {
+  t.plan(2)
+
+  const app = boot()
+
+  app.use((a, b, done) => {
+    t.pass()
+    done(new Error('Kaboom!'))
+  })
+    .use((a, b, done) => t.fail('This should not be called'))
+  app.ready(function (err) {
+    t.error(err)
+  })
+})


delvedor:
I don't agree with this, `avvio` should not decide if is the case to stop the execution.
Let's say you want to load a plugin, and if it fails you want to load another one:
```js
app
  .use(plugin1)
  .after(err => {
     if (err) app.use(plugin2)
  })
```

With your suggestion this won't work.
While you can accomplish your suggestion with:
```js
app
  .use(plugin1)
  .after(err => {
     if (err) throw err
  })
```

delvedor:
Another option could be indeed add an option so support your behavior, something like `failFast: true`.

allevo:
In my case:
```
async function myModule () {
  if (!fastify.mongo) throw new Error('fastify.mongo is falsy')
  // ... other lines
}
fastify
  .register(require('fastify-mongodb'), { ... })
  .register(require('fastify-redis'), { ... })
  .register(myModule)
```

The process fails with `fastify.mongo is falsy` error message.
The debug is hard.

I'm 👍  for `failFast: true` option

allevo:
@delvedor any suggestion for that implementation?

delvedor:
I think that the fastest way to accomplish this is to add something like
```js
if (err != undefined && opts.failFast === true) throw err
```
Here:
https://github.com/mcollina/avvio/blob/b4e2fab72679c5b1a643ce2c2a7fc4bff8f361f5/plugin.js#L38-L48

But I should test it. As you know avvio is pretty complex and there are a lot of side effects.

PR #47, opened 2018-02-05
Support async/await with .after

mcollina:


boot.js
```diff
@@ -273,16 +273,25 @@ function noop () {}
...
+  var res
 
   // with this the error will appear just in the next after/ready callback
   this._error = null
   if (func.length === 0) {
     this._error = err
-    func()
-    process.nextTick(cb)
+    res = func()
+    if (res && res.then) {
```
delvedor: `typeof res.then === 'function'`
Same below

mcollina:
PTAL

PR #48, opened 2018-02-06
fail fast until an after is found

mcollina:
Fixes https://github.com/mcollina/avvio/pull/45
Fixes https://github.com/fastify/fastify/issues/757

cc @fastify @allevo

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,10 @@ app.ready(function (err) {
 })
 ```
 
+When an error happens, the loading of plugins will stop until there is
+an [`after`](#after) callback specified. Otherwise, it will be handled
+in [`ready`](#ready).
+
 -------------------------------------------------------
 <a name="after"></a>
 
@@ -190,6 +194,9 @@ The callback changes basing on the parameters your are giving:
 2. If one parameter is given to the callback, that parameter will be the `error` object.
 3. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.
 4. If three parameters are given to the callback, the first will be the `error` object, the second will be the top level `context` unless you have specified both server and override, in that case the `context` will be what the override returns, and the third the `done` callback.
+
+In the no parameter and one parameter variants, the callback can also
+return a `Promise`.
 
 ```js
 const server = {}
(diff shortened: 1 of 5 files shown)


sarneeh:
Nice! 👍

allevo:
@mcollina

This is a breaking change. Should we ship this on fastify after its rc?
This is not configurable. Is it a problem? 

Anyway it is fine for me

mcollina:
I see this as a bug, not as a breaking change. This behavior was not really documented before, and I would guess it is what people expects. I will update the docs to reflect it, too.

boot.js
```diff
@@ -191,12 +191,13 @@ Boot.prototype._addPlugin = function (plugin, opts) {
   current.enqueue(obj, (err) => {
     if (err) {
       this._error = err
+      // current.error()
```
delvedor: What is this?

jimmiehansson:
Great work and fast solution 🎖 🚀 🤖 🌮

mcollina:
@delvedor check this out.

mcollina:
> IMO the best thing is that if a plugin throws and there is not an after just after it then we should throw.
Since now we have this.isAfter it should be easy to achieve.

`.ready(function (err) { })` expects the errors to end up there. I think that is preferred to a uncatchable throw.

README.md
```diff
@@ -191,6 +195,9 @@ The callback changes basing on the parameters your are giving:
 3. If two parameters are given to the callback, the first will be the `error` object, the second will be the `done` callback.
 4. If three parameters are given to the callback, the first will be the `error` object, the second will be the top level `context` unless you have specified both server and override, in that case the `context` will be what the override returns, and the third the `done` callback.
 
+In the no parameter and one parameter variants, the callback can also
```
jsumners: Add quotes (`"`) around `no parameter` and `one parameter`. It'll be easier to read.

delvedor:
> `ready(function (err) { })` expects the errors to end up there. I think that is preferred to a uncatchable throw.

Yeah, what I mean is:

```js
app.use((instance, opts, next) => {
  next(new Error('kaboom'))
})

app.after(err => {
  console.log(err)
})

app.use((instance, opts, next) => {
  next(new Error('kaboom'))
})

app.use((instance, opts, next) => {
  console.log('this should not be called')
})

app.ready(err => {
  console.log(err)
})
```

jsumners:
So that I am clear:

```js
app
  .use((instance, opts, next) => {
    // Do something
    next()
  })
  .use((instance, opts, next) => {
    next(Error('nope'))
  })
  .use((instance, opts, next) => {
    // skipped
  })

app.ready((err) => {
  // `err` is not `undefined`
  // And the error was not thrown due to a missing `.after`
})

mcollina:
@jsumners yes.

The error is never thrown, it ends up in `ready((err) => {})`.

jsumners:
LGTM with that documentation nit.

PR #49, opened 2018-02-06
preReady event

delvedor:
As titled :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ Options:
 Events:
 
 * `'start'`  when the application starts
+* `'preReady'` fired before the ready queue is run
 
 The `avvio` function can be used also as a
 constructor to inherits from.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -115,6 +115,7 @@ Options:
 Events:
 
 * `'start'`  when the application starts
+* `'preReady'`  fired before `ready` runs
```
mcollina: "fired before the ready queue is run"

boot.js
```diff
@@ -131,6 +131,7 @@ function Boot (server, opts, done) {
 
   Plugin.loadPlugin.call(this, main, (err) => {
     debug('root plugin ready')
+    this.emit('preReady')
```
mcollina: Should we run this if we have errors? Can you add some unit tests to specify that case as well?

delvedor: Since is *pre* I would say yes. I'll add some test.

PR #50, opened 2018-02-16
Catch exceptions in parent.override

jsumners:
This PR should allow https://github.com/fastify/fastify/pull/778 to be fixed.

N.B.: I think `Boot.prototype.start` should return `this`. Do you?

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -36,7 +36,7 @@ Plugin.prototype.exec = function (server, cb) {
   try {
     this.server = this.parent.override(server, func, this.opts)
   } catch (err) {
-    debug('exec errored', name)
+    debug('override errored', name)
     return cb(err)
   }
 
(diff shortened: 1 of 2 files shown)


plugin.js
```diff
@@ -33,7 +33,11 @@ Plugin.prototype.exec = function (server, cb) {
     return
   }
 
-  this.server = this.parent.override(server, func, this.opts)
+  try {
+    this.server = this.parent.override(server, func, this.opts)
+  } catch (err) {
+    return cb(err)
```
delvedor: Can you also add a `debug` log?

mcollina:
> N.B.: I think Boo.prototype.start should return this. Do you?

Yes, can you send another PR?

plugin.js
```diff
@@ -33,7 +33,12 @@ Plugin.prototype.exec = function (server, cb) {
     return
   }
 
-  this.server = this.parent.override(server, func, this.opts)
+  try {
+    this.server = this.parent.override(server, func, this.opts)
+  } catch (err) {
+    debug('exec errored', name)
```
delvedor: I think is more correct `override errored` :D

PR #62, opened 2018-04-21
wrapped close should return a promise with no callback

mcollina:
This fixes two bugs:

1. `await fastify.close()` actually waits for close to happen
2. stop errors from bubbling up if they happen when closing (they will end up in the rejection handler).

cc @StarpTech @allevo

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -58,13 +58,14 @@ function wrap (server, opts, instance) {
     if (func && typeof func !== 'function') {
       throw new Error('not a function')
     }
+
     if (func) {
       instance.close(encapsulateThreeParam(func, this))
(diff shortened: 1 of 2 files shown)


boot.js
```diff
@@ -58,8 +58,13 @@ function wrap (server, opts, instance) {
     if (func && typeof func !== 'function') {
       throw new Error('not a function')
     }
-    instance.close(encapsulateThreeParam(func, this))
-    return this
+    if (func) {
+      instance.close(encapsulateThreeParam(func, this))
+      return this
+    } else {
```
StarpTech: `else` can be removed.

PR #65, opened 2018-06-05
Add Typescript definition

HelloEdit:
Hello

Here is my second attempt to write Typescript definitions for this project.

Already, the main problem is that Typescript cannot handle changing the `expose` option, knowing that they are undefined before the runtime. It is always possible to change this by copying the definitions into an `index.d.ts` file at the root of your project, which will contain the following:
```ts
declare module "avvio" {
    // insert definitions here by changing `use`, `after` and `ready` values.
}
```

So I'd like you to check the function feedback values, just to be sure. You can see that from the test files.

Also to be able to create the right definitions, I introduced the type `avvio.context`. This Type takes relies on conditional typing. If an object is passed as instance in avvio, typescript will add the use, after and ready to the instance functions. If nothing is passed (`null`), the context will be the Boot.

Feel free to ask me questions about this PR ^^

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "Asynchronous bootstrapping of Node applications",
   "main": "boot.js",
   "scripts": {
-    "test": "standard && tap -j4 test/*test.js"
+    "typescript": "tsc --project ./test/types/tsconfig.json",
+    "test": "standard && tap -j4 test/*test.js && npm run typescript"
   },
   "precommit": "test",
   "repository": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -30,11 +31,13 @@
...
   "devDependencies": {
+    "@types/node": "^10.3.0",
     "express": "^4.16.3",
     "pre-commit": "^1.2.2",
     "standard": "^11.0.0",
     "tap": "^11.1.3",
-    "then-sleep": "^1.0.1"
+    "then-sleep": "^1.0.1",
+    "typescript": "^2.9.1"
   },
```
StarpTech: No need to reference the types? https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

HelloEdit: index.d.ts is the default file for typescript definition

StarpTech: Good to know, thanks.

PR #66, opened 2018-06-07
Add 'wrap' option

HelloEdit:
Hello,

This PR is dedicated to adding a flag named `wrap` in the avvio options.

In some cases, avvio should not mutate the instance, but simply reuse it internally. it would allow to have a 100% predictive behavior with Typescript, but also to leave the choice to the developer of the organization of its object (with or without `use`, `after` and `ready`).

Feel free to ask questions about this PR ^^

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ async function third (instance, opts) {
 
 Starts the avvio sequence.
 As the name suggest, `instance` is the object representing your application.
-Avvio will add the functions `use`, `after` and `ready` to the instance.
+Avvio will add the functions `use`, `after` and `ready` to the instance (see options below).
 
 ```js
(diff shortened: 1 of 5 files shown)


HelloEdit:
I just realized that my change does not take into account the server used in function use etc.... 

From my point of view, if wrap is false, Avvio should not mutate the instance in `avvio(instance)`, but use the wraped server internally. I will make a modification of the code and the test to take into account this case.

HelloEdit:
After reflection, I don't think this change as such is useful: it implies some problem and removes the primary utility of Avvio. For now, I prefer to close the PR and maybe reopen it if necessary. Sorry for the inconvenience.

PR #67, opened 2018-06-12
Modularization of Avvio's methods

HelloEdit:
This pull request allows to improve the overload of typecript definitions when the expose option is changed by creating an interface for ready, use and after. I would publish an example in commentary of this PR.

Thanks ^^

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -29,18 +29,9 @@ declare namespace avvio {
   }
 
   interface Server<I> {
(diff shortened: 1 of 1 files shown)


mcollina:
What does it add? Can you add a unit test for this change?

HelloEdit:
it just changes the way you organize Typescript definitions for more modularity with Avvio. The current test verifies this change.

PR #70, opened 2018-08-03
Add promise support for onClose handlers. fixes #69

### LINKED ISSUE #69: async function as onClose handler
If I register an asynchronous onClose handler as a two argument (context, done) function:

```javascript
async function myPlugin (instance, opts) {
  instance.onClose((context, done) => {
    setTimeout(done, 3000)
  })
}
```
when app.close() is called, the next onClose handler does not start until the done() callback is not called on the previous one.

But if I register an asynchronous onClose handler as an async function:

```javascript
async function myPlugin (instance, opts) {
  instance.onClose(async () => {
    await new Promise(resolve => setTimeout(resolve, 3000))
  })
}
```
when app.close() is called, the next onClose handler does not wait until the async function finishes, i.e. the promise resolves.

Is it intentional? It would be much easier to write asynchronous onClose handlers with async functions.

gyszalai:


gyszalai:
oops, I forgot the tests.

gyszalai:
@delvedor please review the tests. I'm not sure if they are correct this way.

test/close.test.js
```diff
@@ -404,3 +404,31 @@ test('close without a cb returns a promise when attaching to a server', (t) => {
...
+test('close with async onClose handlers', (t) => {
+  t.plan(5)
+
+  const app = boot()
+  var order = [1, 2, 3, 4]
+
+  app.onClose(async () => {
+    await new Promise(resolve => setTimeout(resolve, 500))
+    t.is(order.shift(), 3)
+  })
```
mcollina: Can you use Promises instead? We still support Node 6 and Node 4.

gyszalai:
hm... so the tests must be ES5, no async keyword

HelloEdit:
Do we need to update Typescript definition ?

mcollina:
Yes please
Il giorno ven 3 ago 2018 alle 20:04 HelloEdit <notifications@github.com> ha
scritto:

PR #72, opened 2018-08-23
Plugin timeout

mcollina:
This would help solving one of the major pain points of using Fastify, developers not calling their callback and having trouble to debug where the the loading process have stopped.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -99,6 +99,7 @@ function Boot (server, opts, done) {
 
   server = server || this
 
+  this._timeout = Number(opts.timeout) || 0
   this._server = server
   this._current = []
   this._error = null
--- a/boot.js
+++ b/boot.js
@@ -130,12 +130,12 @@ function Boot (server, opts, done) {
   }
 
   this._doStart = null
-  const main = new Plugin(this, (s, opts, done) => {
+  const main = new Plugin(this, function root (s, opts, done) {
     this._doStart = done
     if (opts.autostart) {
       this.start()
     }
-  }, opts, noop)
+  }.bind(this), opts, noop)
 
   Plugin.loadPlugin.call(this, main, (err) => {
     debug('root plugin ready')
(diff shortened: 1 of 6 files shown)


mcollina:
cc @davidmarkclements

davidmarkclements:
yes - I can attest this caused some issues

does the timeout have to be zero by default? 

I suppose so, since it depends on plugins for acceptable timeouts.

But what if the default time out was 500ms, and plugins could define their own timeout to override?

boot.js
```diff
@@ -130,12 +131,12 @@ function Boot (server, opts, done) {
   }
 
   this._doStart = null
-  const main = new Plugin(this, (s, opts, done) => {
+  const main = new Plugin(this, function root (s, opts, done) {
```
delvedor: Why did you remove the arrow function?

plugin.js
```diff
@@ -42,6 +63,15 @@ Plugin.prototype.exec = function (server, cb) {
 
   debug('exec', name)
 
+  var timer
+
+  if (this.parent._timeout > 0) {
+    timer = setTimeout(function () {
+      timer = null
```
delvedor: Can you add a debug log here?

boot.js
mcollina: So I get `root` as the name in the debug.

plugin.js
```diff
@@ -42,6 +63,15 @@ Plugin.prototype.exec = function (server, cb) {
...
+      done(new Error('plugin did not start in time: ' + name))
```
delvedor: I would rephrase this as `plugin '${name}' timed out.`

mcollina: I didn't do that because name can be quite a long string (the first two lines of the function in the bad case). Having those in the middle would make the error unreadable.

davidmarkclements:
Looking awesome 👏

mcollina:
Done.

PR #74, opened 2018-09-03
Avoid swallowing errors in .after() without argument

mcollina:
This is going to be contentious about the semversiness.

Without this change, the following fails:

```js
  const fastify = fastify()
  const payload = { hello: 'world' }

  fastify.register((instance, opts) => {
    return promise.reject(new error('kaboom'))
  }).after(function () {
    t.pass('after is called')
  })

  fastify.inject({
    method: 'get',
    url: '/'
  }).then(() => {
    t.fail('this should not be called')
  }).catch(err => {
    t.ok(err)
    t.strictequal(err.message, 'kaboom')
  })
```

Note that this could cause several _horrible_ bugs like having 404 with routes not defined because an error got swallowed.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -364,7 +364,10 @@ function encapsulateThreeParam (func, that) {
   function _encapsulateThreeParam (err, cb) {
     if (!func) {
       process.nextTick(cb)
-    } else if (func.length === 0 || func.length === 1) {
+    } else if (func.length === 0) {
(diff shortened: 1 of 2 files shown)


mcollina:
that is not affected, as it has 2 arguments.

mcollina:
Included in #75.

PR #75, opened 2018-09-03
 Proper support for after with a passed async function in wrapped mode

mcollina:
Unfortunately `after(async function () {})` did not work properly when wrapping like in Fastify.

Includes #74 as well.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -364,7 +364,10 @@ function encapsulateThreeParam (func, that) {
   function _encapsulateThreeParam (err, cb) {
     if (!func) {
       process.nextTick(cb)
-    } else if (func.length === 0 || func.length === 1) {
+    } else if (func.length === 0) {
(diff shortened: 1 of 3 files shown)


boot.js
```diff
@@ -362,11 +362,27 @@ function encapsulateTwoParam (func, that) {
...
     if (!func) {
       process.nextTick(cb)
-    } else if (func.length === 0 || func.length === 1) {
-      func(err)
-      process.nextTick(cb)
+    } else if (func.length === 0) {
+      res = func()
+      if (res && res.then) {
+        res.then(function () {
+          process.nextTick(cb, err)
```
delvedor: Why do we need `process.nextTick`?

mcollina: to avoid any thrown value within the callback to be catched by the promiss.

PR #76, opened 2018-09-09
Support on close without the instance object

cemremengu:
Attempting https://github.com/fastify/fastify/issues/629

Not sure if this is the right thing to do but based on `encapsulateThreeParam` I came up with this solution.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -354,7 +354,7 @@ function encapsulateTwoParam (func, that) {
       func()
       process.nextTick(cb)
     } else if (func.length === 1) {
-      func(cb)
+      func(this)
       process.nextTick(cb)
     } else {
       func(this, cb)
--- a/boot.js
+++ b/boot.js
@@ -354,8 +354,7 @@ function encapsulateTwoParam (func, that) {
       func()
       process.nextTick(cb)
     } else if (func.length === 1) {
-      func(this)
-      process.nextTick(cb)
+      func(cb)
     } else {
       func(this, cb)
     }
--- a/boot.js
+++ b/boot.js
@@ -351,7 +351,7 @@ function encapsulateTwoParam (func, that) {
   return _encapsulateTwoParam.bind(that)
   function _encapsulateTwoParam (context, cb) {
     if (func.length === 0) {
-      func(this)
+      func()
       process.nextTick(cb)
     } else if (func.length === 1) {
       func(cb)
--- a/boot.js
+++ b/boot.js
@@ -283,7 +283,7 @@ Boot.prototype.ready = function (func) {
   })
 }
 
-function noop () {}
+function noop () { }
 
 function callWithCbOrNextTick (func, cb, context) {
   context = this._server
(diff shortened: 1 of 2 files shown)


mcollina:
Can you add a unit test?

cemremengu:
How about this? I am new to this project so I am kind of hacking around 😄

boot.js
```diff
@@ -350,9 +350,12 @@ function closeWithCbOrNextTick (func, cb, context) {
 function encapsulateTwoParam (func, that) {
   return _encapsulateTwoParam.bind(that)
   function _encapsulateTwoParam (context, cb) {
-    if (func.length === 0 || func.length === 1) {
+    if (func.length === 0) {
       func(this)
```
mcollina: if it is `0` there is no point in calling it with `this` as first argument.

mcollina:
This is a semver-major change

test/close.test.js
```diff
@@ -328,6 +328,20 @@ test('close without a cb', (t) => {
...
+  t.plan(2)
+
+  const app = boot()
+
+  app.onClose((done) => {
+    t.ok('called')
+  })
+
+  app.close(() => {
+    t.pass('closed')
```
mcollina: you might want to:

1. use a `function ( ) {}` so you have access to the `arguments` object
2. place an assertion that checks that `arguments.length === 0`

cemremengu: Looks like length is always 1 as `{0 : null}`  so I can check like:

```js
    t.is(arguments[0], null)
    t.is(arguments[1], undefined)
```

Is that alright?

mcollina: then there is possibly something off with your code, as https://github.com/mcollina/avvio/pull/76/files#diff-fe60a8aed4b3103661f9d88361346b0dR353 should be triggered.

boot.js
```diff
@@ -350,8 +350,11 @@ function closeWithCbOrNextTick (func, cb, context) {
 function encapsulateTwoParam (func, that) {
   return _encapsulateTwoParam.bind(that)
   function _encapsulateTwoParam (context, cb) {
-    if (func.length === 0 || func.length === 1) {
-      func(this)
+    if (func.length === 0) {
+      func()
+      process.nextTick(cb)
+    } else if (func.length === 1) {
+      func(cb)
```
mcollina: You are changing the behavior here. Can you restore `this`?

cemremengu: I did that because of the original issue in fastify stated 

>if onClose(func) is called with a function with 1 parameter, pass the callback and not the >instance as the first one. 

Restored now

cemremengu:
Updated tests to cover that behavior

boot.js
```diff
@@ -350,7 +350,10 @@ function closeWithCbOrNextTick (func, cb, context) {
 function encapsulateTwoParam (func, that) {
   return _encapsulateTwoParam.bind(that)
   function _encapsulateTwoParam (context, cb) {
-    if (func.length === 0 || func.length === 1) {
+    if (func.length === 0) {
+      func()
+      process.nextTick(cb)
+    } else if (func.length === 1) {
       func(this)
       process.nextTick(cb)
```
mcollina: You were right. This needs to change to:

```js
func(cb)
```

Remove the ` process.nextTick` thing.

cemremengu: Tests are failing with that change `not ok child test left in queue:`
```js
function encapsulateTwoParam (func, that) {
  return _encapsulateTwoParam.bind(that)
  function _encapsulateTwoParam (context, cb) {
    if (func.length === 0) {
      func()
      process.nextTick(cb)
    } else if (func.length === 1) {
      func(cb)
    } else {
      func(this, cb)
    }
  }
}
```

cemremengu: Maybe tests need to be updated to call `done()` ?

mcollina:
Yes!
Il giorno lun 10 set 2018 alle 19:50 Cemre Mengu <notifications@github.com>
ha scritto:

Issue #64, opened 2018-05-31
Pretty Print Plugin Tree


StarpTech:
Hi @mcollina, @smartiniOnGitHub has a neat idea to illustrate the plugin tree for diagnostic purposes as in [find-my-way](https://github.com/delvedor/find-my-way#prettyprint) WDYT?

e.g

```md
console.log(avvio.prettyPrint())
// └── /
//   ├── root (fastify v1.0.0) 44ms
//   │   └── /fastify-auth (v0.0.4) 2ms
//   │   └── /fastify-swagger (v0.4.4) 23ms
//   │   └── /fastify-custom (v1.0.0) 19ms
//   │   	└── /fastify-a (v1.0.0) 9ms
//   │   	└── /fastify-b (v1.0.0) 10ms
```
The tree would illustrate the loading order and other useful informations like plugin `name`, `version`, `loading-time`. Dependencies across different plugins could be marked in different colors. There are many options 😄.

### `avvio.prettyPrint([updates=false])` 
Should be able to call it before everything is bootsrapped. The cli can [update](https://github.com/sindresorhus/log-update) itself so it's possible to see which plugin has issues to bootstrap.

mcollina:
Note that avvio does not anything about prefixes, those are fastify thing.
Other than that, *go for it*.

delvedor:
I like it!
I suggest to use [node-archy](https://github.com/substack/node-archy), in `find-my-way` we are using a way to complex method :P

We should think how to handle unnamed plugins.

StarpTech:
Good note. This must be enriched in a different way. I leave it open so anybody can start a try.

climba03003:
Resolved in #81

PR #81, opened 2019-02-13
feature/pretty print

Eomm:
Hi, this PR would try to implement #64 

Now the status, applying the pretty print on the example.js:

```
avvio
└─┬ boot undefined ms
  ├── bound root 8 ms
  ├── first 5 ms
  ├── second 1 ms
  ├── bound _after 1 ms
  └── third 0 ms
```

Looking the cose, I don't realize why the `parent`, take in Plugin with `this.nameParent = this.parent._current.name` is always undefined.

I'm missing something but I can't figure out what 😅
Do you have any tips to help me?

_tests will be added_

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ async function third (instance, opts) {
   * <a href="#onClose"><code>instance.<b>onClose()</b></code></a>
   * <a href="#close"><code>instance.<b>close()</b></code></a>
   * <a href="#express"><code>avvio.<b>express()</b></code></a>
+  * <a href="#toJSON"><code>avvio.<b>toJSON()</b></code></a>
+  * <a href="#prettyPrint"><code>avvio.<b>prettyPrint()</b></code></a>
 
 -------------------------------------------------------
 <a name="constructor"></a>
@@ -472,6 +474,73 @@ app.close()
 ```
 
 `done` must be called only once.
+
+-------------------------------------------------------
+
+<a name="toJSON"></a>
+
+### avvio.toJSON()
+
+Return a JSON tree rappresenting the state of the plugins and the loading time.
+Call it on `preReady` to get the complete tree.
+
+```js
+const avvio = require('avvio')()
+avvio.on('preReady', () => {
+  avvio.toJSON()
+})
+```
+
+The ouput is like this:
+```json
+{
+   "avvio":{
+      "label":"avvio",
+      "start":1550165276637,
+      "stop":1550165276639,
+      "diff":2,
+      "children":[
+         {
+            "label":"one",
+            "start":1550165276639,
+            "stop":1550165276639,
+            "diff":0
+         },
+         {
+            "label":"two",
+            "start":1550165276639,
+            "stop":1550165276639,
+            "diff":0
+         }
+      ]
+   }
+}
+```
+
+-------------------------------------------------------
+
+<a name="prettyPrint"></a>
+
+### avvio.prettyPrint()
+
+Print to `console.log` the tree returned by the `toJSON()` method.
+
+```js
+const avvio = require('avvio')()
+avvio.on('preReady', () => {
+  avvio.prettyPrint()
+})
+```
+
+The ouput will be like:
+
+```
+avvio 56 ms
+├── first 52 ms
+├── second 1 ms
+├── bound _after 0 ms
+└── third 2 ms
+```
 
 -------------------------------------------------------
 
--- a/README.md
+++ b/README.md
@@ -494,26 +494,45 @@ avvio.on('preReady', () => {
 The ouput is like this:
 ```json
 {
-   "avvio":{
-      "label":"avvio",
-      "start":1550165276637,
-      "stop":1550165276639,
-      "diff":2,
-      "children":[
-         {
-            "label":"one",
-            "start":1550165276639,
-            "stop":1550165276639,
-            "diff":0
-         },
-         {
-            "label":"two",
-            "start":1550165276639,
-            "stop":1550165276639,
-            "diff":0
-         }
-      ]
-   }
+  "label": "bound root",
+  "start": 1550245184665,
+  "nodes": [
+    {
+      "parent": "bound root",
+      "start": 1550245184665,
+      "label": "first",
+      "nodes": [
+        {
+          "parent": "first",
+          "start": 1550245184708,
+          "label": "second",
+          "nodes": [],
+          "stop": 1550245184709,
+          "diff": 1
+        }
+      ],
+      "stop": 1550245184709,
+      "diff": 44
+    },
+    {
+      "parent": "bound root",
+      "start": 1550245184709,
+      "label": "bound _after",
+      "nodes": [],
+      "stop": 1550245184709,
+      "diff": 0
+    },
+    {
+      "parent": "bound root",
+      "start": 1550245184709,
+      "label": "third",
+      "nodes": [],
+      "stop": 1550245184709,
+      "diff": 0
+    }
+  ],
+  "stop": 1550245184709,
+  "diff": 44
 }
 ```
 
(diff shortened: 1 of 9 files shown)


mcollina:
I don’t understand your question 🤣. Can you make an example?

Overall I would prefer if we add a way to get a JS/JSON representation than something with archy.

Eomm:
Sorry, I needed only a rubber duck debug 🙌
I have added test for the `toJSON` function.

Would you like to remove the `prettyPrint` function?

test/to-json.test.js
```diff
@@ -0,0 +1,63 @@
...
+    outJson.avvio.children.push({
+      label: 'three',
+      start: /\d*/,
+      stop: /\d*/,
+      diff: /\d*/
+    })
+    t.like(json, outJson)
+    done()
+  }
+})
```
mcollina: Can you add another test with a multi-level hierarchy of plugins?

Eomm:
> What will this print and produce for https://github.com/fastify/fastify/blob/master/examples/example.js?
> 
> Actually, let's keep the archy part, as it's nice and useful.

The complete output is:
```
$ node example.js
first loaded { hello: 'world' }
second loaded
after first and second
third loaded
avvio 53 ms
├── first 50 ms
├── second 0 ms
├── bound _after 1 ms
└── third 0 ms

application booted!
```

mcollina:
I think we are talking about two different examples, I was referring to the fastify one.

test/to-json.test.js
Eomm: I'm not sure: what do you expect?
Because I get the same output as a single level: in fact, the plugins are all loaded sequentially

mcollina: I think the goal should be to show a tree, not just the loading order.

mcollina: In other terms, the information that two was loaded from one is important.

Eomm:
Yeah, here you are:
```
~/workspace/fastify (master)
$ node examples/example.js
avvio 13 ms
└── bound _after 0 ms

server listening on 3000
```

I'll check what is going on 🔍

test/to-json.test.js
Eomm: fixed
```
bound root 3 ms
└─┬ first 3 ms
  ├── second 0 ms
  └─┬ third 2 ms
    └── fourth 1 ms
```

But I have to check way in fastify/example I get:

```$ node examples/example.js
bound root 11 ms
├── bound _after 5 ms
├── bound _after undefined ms
├── bound _after undefined ms
├── bound _after undefined ms
├── bound _after undefined ms
├── bound _after undefined ms
├── bound _after undefined ms
├── bound _after undefined ms
└── bound _after undefined ms
```

mcollina: That is more or less what it should be happening :).
Here is the culprit: https://github.com/fastify/fastify/blob/356cf6f078a9f93b009ddb6384deaffe549d1032/fastify.js#L589.

I don't know how we would like to represent those.

Eomm:
I have fixed the last things. Now running the fastify/example I get a coherent results (before I was getting undefined for the times):

Let me know if you think that there are other fixes to add.

Thank you

```
$ node examples/example.js
bound root 11 ms
├── bound _after 3 ms
├── bound _after 1 ms
├── bound _after 0 ms
├── bound _after 0 ms
├── bound _after 0 ms
├── bound _after 0 ms
├── bound _after 0 ms
├── bound _after 0 ms
└── bound _after 0 ms

server listening on 3000
```

mcollina:
How can we skip `bound _after 3 ms`?

Specifically, we use it in fastify _for each defined route_: https://github.com/fastify/fastify/blob/1313bea84826d3bb2d09857f149fcb7f38c4de7b/fastify.js#L589-L600. We definitely do not want that to show up.

time-tree.js
```diff
@@ -0,0 +1,86 @@
...
+TimeTree.prototype.getParent = function (parent, nodeId) {
+  if (parent === null) {
+    return this.root
+  }
+
+  return this.search(node => node.label === parent)
+}
+
+TimeTree.prototype.getNode = function (nodeId) {
+  return this.search(node => node.id === nodeId)
```
mcollina: This is going to get pretty hot, and slow us during boot time. Can you create a Map where you store the ids and look them up instead?

Eomm:
> We definitely do not want that to show up.

Done. I didn't do it configurable since it is not worth seeing the `after` if is ok for you.

Now an example with `fastify.register` produce:
```
  fastify.register((i, a, n)=>{
    i.register( (i, a, n)=>{
      setTimeout(n, 100)
    })
    i.register((i, a, n)=>{
      setTimeout(n, 200)
    })
    i.register((i, a, n)=>{
      setTimeout(n, 300)
    })
    i.register((i, a, n)=>{
      setTimeout(n, 400)
    })
    setTimeout(n, 500)
  })



bound root 1518 ms
└─┬ (i, a, n)=>{ -- i.register((i, a, n)=>{ 1509 ms
  ├── (i, a, n)=>{ -- setTimeout(n, 100) 101 ms
  ├── (i, a, n)=>{ -- setTimeout(n, 200) 203 ms
  ├── (i, a, n)=>{ -- setTimeout(n, 300) 300 ms
  └── (i, a, n)=>{ -- setTimeout(n, 400) 401 ms
```

And the examples/example show only:
```
$ node examples/example.js
bound root 20 ms

server listening on 3000
```

mcollina:
Is this stuck on me? :D

Eomm:
Nope, I didn't see the comment to introduce a Map, I'm fixing 🙌

Eomm:
@mcollina I have completed the changes

Now example will print out as expected:

```
bound root 190 ms
├─┬ first 50 ms
│ └── second 1 ms
├── duplicate 22 ms
├─┬ duplicate 115 ms
│ └─┬ duplicate 94 ms
│   └─┬ duplicate 62 ms
│     └─┬ duplicate 41 ms
│       └── duplicate 20 ms
└── third 0 ms
```

README.md
```diff
@@ -475,6 +477,83 @@ app.close()
...
+Print to `console.log` the tree returned by the `toJSON()` method.
+
+```js
+const avvio = require('avvio')()
+avvio.on('preReady', () => {
+  avvio.prettyPrint()
+})
+```
+
+The ouput will be like:
```
delvedor:
```suggestion
The output will be like:
```

time-tree.js
```diff
@@ -0,0 +1,88 @@
...
+TimeTree.prototype.prittyPrint = function () {
+  const decorateText = (node) => {
+    node.label = `${node.label} ${node.diff} ms`
+    if (node.nodes.length > 0) {
+      node.nodes = node.nodes.map(_ => decorateText(_))
+    }
+    return node
+  }
+  const out = decorateText(this.toJSON())
+  console.log(archy(out))
```
delvedor: I think that `prettyPrint` should return a string instead of print it.
It should also make it easier to test.

Eomm: Done

mcollina:
CI is failing  on Node 8:

```
test/pretty-print.test.js ........................... 13/14
  pretty print
  not ok should match pattern provided
    found: ├── first 12 ms
    pattern: '/├── first \d{1} ms/'
    at:
      line: 37
      column: 9
      file: test/pretty-print.test.js
      function: print.split.forEach
    stack: |
      print.split.forEach (test/pretty-print.test.js:37:9)
      Array.forEach (<anonymous>)
      Boot.show (test/pretty-print.test.js:36:23)
      Boot.Plugin.loadPlugin.call (boot.js:147:10)
      toLoad.finish (plugin.js:178:7)
      done (plugin.js:136:5)
      check (plugin.js:147:7)
    source: |
      t.match(l, linesExpected[i])
```

Eomm:
Fixed, I was too much optimistic... Now I understand why under-pressure has the CI failing every time 😂

PR #82, opened 2019-02-27
Added support for a function in place of opts

alex-ppg:
Attempting [fastify/fastify#1489](https://github.com/fastify/fastify/issues/1489)

An external dependency `fast-deepclone` was added as `Object.assign({}, variable)` wouldn't suit our purpose due to it being a shallow copy and thus passing by reference. The operational overhead is minimal thanks to the module being coded natively in C++. Its license is [MIT](https://github.com/scottinet/fast-deepclone/blob/master/LICENSE) which I believe is compatibly with this project's license.

Added unit tests as well. Let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,22 @@
 # avvio
 
 [![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat)](http://standardjs.com/)
- [![Build Status](https://travis-ci.org/mcollina/avvio.svg)](https://travis-ci.org/mcollina/avvio)
+[![Build Status](https://travis-ci.org/mcollina/avvio.svg)](https://travis-ci.org/mcollina/avvio)
 
-Asynchronous bootstrapping is hard, different things can go wrong, *error handling* and *load order* just to name a few. The aim of this module is to make it simple.
+Asynchronous bootstrapping is hard, different things can go wrong, _error handling_ and _load order_ just to name a few. The aim of this module is to make it simple.
 
-`avvio` is fully *reentrant* and *graph-based*. You can load
-components/plugins *within* plugins, and be still sure that things will
+`avvio` is fully _reentrant_ and _graph-based_. You can load
+components/plugins _within_ plugins, and be still sure that things will
 happen in the right order. At the end of the loading, your application will start.
 
-* [Install](#install)
-* [Example](#example)
-* [API](#api)
-* [Acknowledgements](#acknowledgements)
-* [License](#license)
+- [Install](#install)
+- [Example](#example)
+- [API](#api)
+- [Acknowledgements](#acknowledgements)
+- [License](#license)
 
 <a name="install"></a>
+
 ## Install
 
 To install `avvio`, simply use npm:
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -41,6 +41,7 @@
   },
   "dependencies": {
     "debug": "^4.0.0",
+    "fast-deepclone": "^1.0.1",
```
mcollina: Can you use https://github.com/davidmarkclements/rfdc instead? I prefer no native dependency on this module.

plugin.js
```diff
@@ -25,9 +26,9 @@ function getName (func) {
   return func.toString().split('\n').slice(0, 2).map(s => s.trim()).join(' -- ')
 }
 
-function Plugin (parent, func, opts, isAfter, timeout) {
+function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.func = func
-  this.opts = opts
+  this.opts = isFunction(optsOrFunc) ? deepClone(optsOrFunc()) : optsOrFunc
```
mcollina: Why not?

```suggestion
  this.opts = optsOrFunc instanceof Function ? deepClone(optsOrFunc()) : optsOrFunc
```

alex-ppg: My bad, the segment copied was fixing an issue that only exists in browsers when performing cross-frame checks. I implemented your suggestion and cleaned up `isFunction`.

package.json
alex-ppg: Answering your question above, deep-cloning should be done as unsuspecting users might pass-by-reference internal settings set by another plugin. Take the following snippet into consideration:

first-plugin.js: 
```
const fp = require('fastify-plugin')

module.exports = fp((fastify, options) => {
    // Pseudo-code
    const establishConnection = opts => createConnection(opts)
    // Pseudo-code end
    fastify.decorate('conn_settings', options) 
    fastify.decorate('create_connection_1', establishConnection)
})
```
second-plugin.js:
```
const fp = require('fastify-plugin')

module.exports = fp((fastify, opts) => {
    opts.port = Number.isNaN(opts.port) ? Number(opts.port) : opts.port
    fastify.decorate('my_conn_2', establishConnection(opts))
})
```
```
fastify.register(require('./first-plugin'));
fastify.register(require('./second-plugin'), ()=>fastify.conn_settings);
fastify.ready(err => {
    fastify.create_connection_1() // Error: Expected opts.port to be string but got number
})
```

I know this is not the most ideal example but it can become a danger nonetheless. This is an end-user error so we can ignore this if we wish so.

Sorry for the late responses, today is a busy day.

alex-ppg: The issue above is also indicated in the test sub-case on basic.test.js. I can revert the changes and simply write the initial `optsOrFunc instanceof Function ? optsOrFunc() : optsOrFunc` without the extras if you believe the deep-copy is unnecessary.

plugin.js
```diff
@@ -25,9 +26,9 @@ function getName (func) {
   return func.toString().split('\n').slice(0, 2).map(s => s.trim()).join(' -- ')
 }
 
-function Plugin (parent, func, opts, isAfter, timeout) {
+function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.func = func
-  this.opts = opts
+  this.opts = optsOrFunc instanceof Function ? deepClone(optsOrFunc()) : optsOrFunc
```
delvedor: Can you use `typeof` instead of `instanceof`? Same result, better performances :D
```suggestion
  this.opts = typeof optsOrFunc === 'function' ? deepClone(optsOrFunc()) : optsOrFunc
```

package.json
delvedor: Agree with @mcollina that we should not use native dependencies.
Regarding the deep cloning, I don't think we should introduce it, JavaScript passes objects by reference, and the user should know this.
If someone really needs to clone the object, they can do that by adding the cloning utility inside their function :)

Safe defaults for the win!

mcollina: I agree that removing deep cloning is the safest option here.

plugin.js
alex-ppg: I thought `instanceof` would be faster as it doesn't do string comparison but after testing I was wrong, there is a whopping 410m ops/sec difference for objects and 440m ops/sec for functions. I will integrate your commit in this push along with the removal of deep cloning.

package.json
alex-ppg: I removed the additional test case (as it would fail) and removed the deep-cloning dependency. I should have created a basic commit and asked for enhancements rather than try to path up everything from the get go

plugin.js
```diff
@@ -25,9 +25,9 @@ function getName (func) {
   return func.toString().split('\n').slice(0, 2).map(s => s.trim()).join(' -- ')
 }
 
-function Plugin (parent, func, opts, isAfter, timeout) {
+function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.func = func
-  this.opts = opts
+  this.opts = typeof optsOrFunc === 'function' ? optsOrFunc() : optsOrFunc
```
mcollina:
```suggestion
  this.opts = typeof optsOrFunc === 'function' ? optsOrFunc(parent) : optsOrFunc
```

It might be helpful to call the opts function with the parent fastify instance. It will be handy so opts function could not rely on closures if they decide so.

alex-ppg: I agree, users might want to create local-scope functions. It is a simple change so I have added it along with the documentation in the latest commit.

alex-ppg:
Thanks, added the test as well. Let me know if anything else is missing!

README.md
```diff
@@ -142,11 +142,14 @@ app.on('start', () => {
...
 <a name="use"></a>
 
-### app.use(func, [opts])
+### app.use(func, [optsOrFunc])
 
 Loads one or more functions asynchronously.
+
 The function **must** have the signature: `instance, options, done`
 
+However, if the function is a `Promise` (i.e. `async`), it **must** have the signature: `instance, options`
```
mcollina: This should work also with `use(async function (instance) {})` or `use(async function () {})`. Why do you say it's required?

test/basic.test.js
```diff
@@ -147,6 +147,30 @@ test('boot a plugin with options', (t) => {
...
+  app.use(function (s, opts, done) {
+    t.equal(s, server, 'the first argument is the server')
+    t.deepEqual(opts, myOpts, 'passed options via function')
+    done()
+  }, myOptsAsFunc)
+
+  app.on('start', () => {
+    t.pass('booted')
+  })
+})
```
mcollina: Can you also add a test that demonstrates this sentence:

> This is useful in cases where an injected variable from a plugin needs to be made available to another.

It's the whole point of this change, and we need to verify that we don't regress.

mcollina:
I'm sorry for all the back-and-forth on this, but it's a very important change for Avvio and Fastify.

test/basic.test.js
alex-ppg: Thank you for taking your time to vet the code carefully. The fix you had purposed for passing in the parent variable to the context of the function failed as it was evaluated at the time the `Plugin` was created. I fixed it by utilizing `bind` and executing it during `Plugin.prototype.exec` which fixed the issue.

README.md
alex-ppg: You are correct, I didn't read into it but simply wanted to add a note for `async` functions as the first **must** does not hold in such functions. Fixed it in the latest commit.

alex-ppg:
There was indeed an issue with the way the `parent` reference was passed in to the `optsOrFunc` context. As it was evaluated on creation, it wasn't being passed correctly. I added a test case and fixed the issue with `bind`.

plugin.js
```diff
@@ -25,9 +25,9 @@ function getName (func) {
   return func.toString().split('\n').slice(0, 2).map(s => s.trim()).join(' -- ')
 }
 
-function Plugin (parent, func, opts, isAfter, timeout) {
+function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.func = func
-  this.opts = opts
+  this.opts = typeof optsOrFunc === 'function' ? optsOrFunc.bind(optsOrFunc, parent) : optsOrFunc
```
mcollina:
```suggestion
  this.opts = typeof optsOrFunc === 'function' ? optsOrFunc.bind(null, parent) : optsOrFunc
```

alex-ppg:
I added self-binding in case the original function was already bound to another instance. Anyhow, I accepted your change as I suppose you wish to keep things simple.

README.md
mcollina: This does not seems fixed.

Also, the function is not a Promise, but _returns_ a promise.

alex-ppg: Sorry for wasting your time with minor issues, I fixed it in the latest commit.

mcollina:
Can you please restore the README as it was before? There are a ton of formatting changes.

alex-ppg:
Yeah I usually code with Prettier and had left it on, sorry. It should be back to the original state.

mcollina:
Can you also send a PR to update the fastify docs? Thanks!

alex-ppg:
I would but I already have a fork that has passed preview but has not been merged and I am unable to create a branch due to no-contribution access. I already have the changes locally so when the fork is merged I will simply re-fork and make a PR.

mcollina:
I don't understand.

alex-ppg:
The Fastify PR [#1490](https://github.com/fastify/fastify/pull/1490) is based on my fork of fastify. As I am unable to create a new branch and open a PR, the only way is via forks and it is impossible to have two forks of the same repository.

mcollina:
You can create mulitple branches on a single forked repository and send multiple PRs.

alex-ppg:
Makes sense although I had never read that anywhere, thanks.

PR #88, opened 2019-08-02
Do not crash within after when registering a plugin

mcollina:
This revert a bit done what was done in https://github.com/mcollina/avvio/pull/81 by @Eomm. We have to report the `after` calls to the time tracker, otherwise we lose the trace of what we have loaded. However, this pollutes the `.toJSON()` and pretty-print representation of this.

I think we should land this as patch asap, and then find out what to do with the pretty print.

Fixes: https://github.com/fastify/fastify/issues/1781

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -204,14 +204,12 @@ Boot.prototype._addPlugin = function (plugin, opts, isAfter) {
   const current = this._current[0]
 
   const obj = new Plugin(this, plugin, opts, isAfter)
-  if (!isAfter) {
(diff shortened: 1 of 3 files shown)


Eomm:
Gotcha, I was thinking if we could try perf_hooks? (but it is experimental)

Anyway the main issue here is get a unique name of the function which is always the same with arrow functions

mcollina:
We can use a global counter?

Il giorno ven 2 ago 2019 alle 19:53 Manuel Spigolon <
notifications@github.com> ha scritto:

PR #89, opened 2019-08-05
fix test pretty print after

Eomm:
Since the test wasn't failing due #88, I would add the `after` call in the pretty tests.

This is the output:

```
bound root 195 ms
├── first 0 ms
├─┬ duplicate 86 ms
│ └─┬ duplicate 64 ms
│   └─┬ duplicate 42 ms
│     └── duplicate 21 ms
├── second 0 ms
├─┬ bound _after 0 ms
│ └── afterInsider 0 ms
├── bound _after 0 ms
├─┬ duplicate 63 ms
│ └─┬ duplicate 41 ms
│   └── duplicate 21 ms
├── third 0 ms
├── bound _after 0 ms
└─┬ duplicate 41 ms
  └── duplicate 21 ms
```

that is not good for fastify because there is always a `bound _after` entry for each `register` call.

I will try to fix this behaviour in separate PR

### DIFF
--- a/test/pretty-print.test.js
+++ b/test/pretty-print.test.js
@@ -4,7 +4,7 @@ const test = require('tap').test
 const boot = require('..')
 
 test('pretty print', t => {
-  t.plan(14)
+  t.plan(16)
 
   const app = boot()
   app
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks!!

PR #90, opened 2019-09-08
Add onRegister for plugins

SerayaEryn:
As discussed in fastify/fastify#1812 this adds support for an optional `onRegister` function that can be called synchronously before a plugin is loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,20 @@ app.use(second, func)
 
 This is useful in cases where an injected variable from a plugin needs to be made available to another.
 
+If the meta data of a plugin contain a `onRegister(server, opts)` function, it will be executed synchronous before the plugin is being loaded: 
+
+```js
+async function plugin (app, opts) {
+  console.log(opts.list)
+})
+plugin[Symbol.for('plugin-meta')] = {
+  onRegister (app, opts) {  // this is a sync function, no callback and no promise
+    opts.list = ['foo', 'bar']
+  }
+}
+app.use(plugin)
+```
+
 -------------------------------------------------------
 <a name="error-handling"></a>
 #### Error handling
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -355,9 +355,10 @@ test('boot a promise plugin with onRegister', (t) => {
     hello: 'world'
   }
 
-  async function plugin (s, opts) {
+  function plugin (s, opts) {
     t.equal(s, server, 'the first argument is the server')
     t.deepEqual(opts, myOpts, 'passed options')
+    return Promise.resolve()
   }
   plugin[Symbol.for('plugin-meta')] = {
     onRegister (app, opts) {
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -371,4 +371,3 @@ test('boot a promise plugin with onRegister', (t) => {
     t.pass('booted')
   })
 })
-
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -204,6 +204,20 @@ app.use(second, func)
 
 This is useful in cases where an injected variable from a plugin needs to be made available to another.
 
+If the meta data of a plugin contain a `onRegister(server, opts)` function, it will be executed synchronous before the plugin is being loaded: 
+
+```js
+async function plugin (app, opts) {
+  console.log(opts.list)
+})
+plugin[Symbol.for('plugin-meta')] = {
```
StarpTech: A user could overwrite the `fastify-plugin` meta data when they define it afterwards. I would use a different symbol.

mcollina: While this is true, I think it’s the user responsibility to keep the previous info around when overriding.

I prefer not to add another symbol, as it might become harder to integrate later.

test/basic.test.js
```diff
@@ -318,3 +318,30 @@ test('throws correctly if registering after ready', (t) => {
...
+      t.equal(app, server, 'the first argument is the server')
+      t.deepEqual(opts, myOpts, 'passed options')
+    }
+  }
+  app.use(plugin, myOpts)
+
+  app.on('start', () => {
+    t.pass('booted')
+  })
+})
```
Eomm:
```suggestion
})

test('boot a promise plugin with onRegister', (t) => {
  t.plan(5)

  const server = {}
  const app = boot(server)
  const myOpts = {
    hello: 'world'
  }

  async function plugin (s, opts) {
    t.equal(s, server, 'the first argument is the server')
    t.deepEqual(opts, myOpts, 'passed options')
  }
  plugin[Symbol.for('plugin-meta')] = {
    onRegister (app, opts) {
      t.equal(app, server, 'the first argument is the server')
      t.deepEqual(opts, myOpts, 'passed options')
    }
  }
  app.use(plugin, myOpts)

  app.on('start', () => {
    t.pass('booted')
  })
})

```

test/basic.test.js
```diff
@@ -318,3 +318,57 @@ test('throws correctly if registering after ready', (t) => {
...
+
```
Eomm:
```suggestion
```

Sorry.. lining 😇

SerayaEryn: No problem

SerayaEryn:
> In my opinion, we should think of a different approach to solving the issue.

I am not sure how another approch would look like. But I'd be happy to take a on another approch if there if someone has an idea. :)

Eomm:
> Hello! Thank you for the pr!
> I'm 👎 with this change because while it improves the overall developer experience, makes a quite complicated and intricated plugin developer experience.
> In my opinion, we should think of a different approach to solving the issue.

I would propose this: the plugin can implement the `on` function like event emitter:

```
function plugin (s, opts) {
  console.log(`Hello ${opts.name}`)
  return 'hello'
}

plugin.on = function (event, opts) {
  // edit opts
}
```

So `avvio` should check if it must call the on sync function with some events like:
- onRegister
- onWhatWeWant

I'm not sure if we can also use directly an `EventEmitter` since we are not using `new`:

```
const { EventEmitter } = require('events')
const util = require('util')
util.inherits(plugin, EventEmitter)

function plugin (s, opts, cb) {
  console.log(`Hello ${opts.name}`)
  cb()
}
```

Fdawgs:
Any updates on this? PR is over 3 years old at this point, is it still required?

mcollina:
Can be closed

PR #91, opened 2019-12-01
Basic support for promise plugins / esm with import

mcollina:
The goal of this PR is to add support for `import()` statements for plugins. In this way, if we want to load an ESM, Node could potentially load all of them in parallel while we process one at a time.

In fastify lingo:

```js
app.register(import('path/to/plugin'), opts)
```

The way this is implemented in this PR is to add support for promises: import() returns a promise.

This needs:

* [x] some more testing with latest Node 13
* [x] docs

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -187,7 +187,6 @@ function assertPlugin (plugin) {
 
 // load a plugin
 Boot.prototype.use = function (plugin, opts) {
-  assertPlugin(plugin)
   this._addPlugin(plugin, opts, false)
 
   return this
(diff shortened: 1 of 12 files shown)


boot.js
```diff
@@ -179,21 +179,22 @@ Boot.prototype.override = function (server, func, opts) {
...
+}
+
 // load a plugin
 Boot.prototype.use = function (plugin, opts) {
-  if (typeof plugin === 'function') {
-    this._addPlugin(plugin, opts, false)
-  } else {
-    throw new Error('plugin must be a function')
-  }
+  assertPlugin(plugin)
```
delvedor: There is no need for this check, as we will do it in any case in `_addPlugin`.

mcollina:
@delvedor done! Please check it out.

PR #92, opened 2020-01-09
support await after / await use

davidmarkclements:
for fastify this will enable

```js
async function plugin (fastify, opts) {
  await fastify.register(somePlugin)
  console.log('after somePlugin has registered')
  fastify.register(anotherPlugin)
}
```

as an alternative to 

```js
async function plugin (fastify, opts) {
  fastify.register(somePlugin)
  fastify.after(() => {
    console.log('after somePlugin has registered')
    fastify.register(anotherPlugin)
  }) 
}
```


davidmarkclements:
CI is passing in all node version except for node 6. Does this need to be addressed or can we drop 6?

cc @mcollina @delvedor @jsumners

mcollina:
Please target https://github.com/mcollina/avvio/tree/next instead and drop 6 and 8.

davidmarkclements:
@mcollina done

README.md
```diff
@@ -163,18 +163,23 @@ app.use(plugin)
...
-async/await is also supported:
+
+`use` returns a thenable wrapped instance on which `use` is called, to support a chainable API that can also be awaited.
+
+This way, async/await is also supported and `use` can be awaited instead of using `after`:
 
 ```js
-async function plugin (server, opts) {
-  await sleep(10)
+async function main () {
```
jsumners: Since the lead-in says "instead of using `after`:", I think it would be a good idea to show both examples. Something like:

> Using `after`:
> ```js
> // example
> ```
>
> Using the thenable:
> ```js
> // your current example
> ```

boot.js
```diff
@@ -26,10 +26,26 @@ function wrap (server, opts, instance) {
   if (server[readyKey]) {
     throw new Error(readyKey + '() is already defined, specify an expose option')
   }
-
   server[useKey] = function (a, b, c) {
     instance.use(a, b, c)
-    return this
+    return {
+      __proto__: this,
```
jsumners: Doesn't this need to be `'__proto__'` (https://tc39.es/ecma262/#sec-__proto__-property-names-in-object-initializers)?

And isn't setting the prototype in this way considered deprecated?

mcollina: Using `Object.create()` will achieve the same thing.

jsumners: I think that would be better.

davidmarkclements: As an object key it's already a string, it doesn't have to be explicitly quoted.

The case for `{__proto__}`: `__proto__` in an object literal is an uncontested legitimate part of the spec (as opposed to the `obj.__proto__` getter/setter property which is contested). Personally I think it's more elegant - although I wouldn't use it in application code I would use it in something like this. `Object.create` will create more boilerplate noise (and will need more objects for the descriptor) but if there's consensus on changing it I'll change it np. @mcollina @delvedor if you weigh in on preference in addition to @jsumners 
 I'll act accordingly

README.md
davidmarkclements: good idea :+1:

boot.js
jsumners: > As an object key it's already a string, it doesn't have to be explicitly quoted.

Yeah, but this particular "string" makes me uneasy. I think it is ambiguous as to what the runtime will actually do in this case.

mcollina: I think Object.create() is more idiomatic and readable.

delvedor: I agree with @mcollina.

davidmarkclements: 😆 then I defer to the consensus

davidmarkclements:
> Great work!!
> Can you add a more complex test that verifies what happens when you register multiple plugins, some await and some not?
> For example:
> 
> ```
> await app.use(first)
> // check decorated app
> app.use(second)
> // check that second has noot decorated the instance
> await app.use(third => app.use(fourth))
> // check that we have third and fourth decorators
> await app.ready()
> // check all
> ```

So the way this is currently implemented is the queue is processed when we hit an await, to put this example into `after` terms this is what currently happens:

```js
app.use(first)
app.after(() => {
  // check decorated app
  app.use(second)
  app.use(async function third (app) { app.use(fourth) })
  // CANT CHECK HERE BECAUSE AWAIT WILL TRIGGER AN AFTER
  app.after(async () => {
     // check that second has not decorated the instance (SPOILER, IT HAS)
     // check that we have third and fourth decorators
     await app.ready()
     // check all
  })
})
```

Are we saying it's essential to ensure that non-awaited `use` aren't processed until the end of the plugin function? /cc @mcollina

delvedor:
> Are we saying it's essential to ensure that non-awaited use aren't processed until the end of the plugin function? 

I would not say that is essential, but it's important to at least document what's happening here.
Basically awaiting a `use` is similar to calling `ready` multiple times? Does it load everything that has been added until that point?
I'm trying to understand how should we signal this new feature, and if (and how) it will break existing code :)

mcollina:
> Are we saying it's essential to ensure that non-awaited use aren't processed until the end of the plugin function? 

Yes. Right now, plugins are run after the main function complete. I think we should retain this behavior.
We should not start processing of plugins automatically.

> Basically awaiting a use is similar to calling ready multiple times? Does it load everything that has been added until that point?

I think that should be the goal, yes.

davidmarkclements:
@mcollina wait.. am I misunderstanding or is that a contradiction. 

* non-awaited plugins should run after main function
* loading everything added until that point (the point where an `await` occurs)

I think I agree it's worth the effort to keep the behaviour and unsurprising as possible, which means processing unawaited plugins at the end of the function

delvedor:
We have two options:
**1**
```
app.use(first)
// app.firstDecorator = undefined

await app.use(second)
// app.firstDecorator = true
// app.secondDecorator = true

app.use(third)
// app.thirdDecorator = undefined

await app.ready()
// app.firstDecorator = true
// app.secondDecorator = true
// app.thirdDecorator = true
```

**2**
```
app.use(first)
// app.firstDecorator = undefined

await app.use(second)
// app.firstDecorator = undefined
// app.secondDecorator = true

app.use(third)
// app.thirdDecorator = undefined

await app.ready()
// app.firstDecorator = true
// app.secondDecorator = true
// app.thirdDecorator = true
```

Which one do you think we should pick?
The option **2** is the one that does not introduce breaking changes with the current behavior.

davidmarkclements:
the non-breaking changes one (2) - it's just a case of making it work

mcollina:
Ahum, I think option 1 is better and it's compatible to how .after() wroks right now.

davidmarkclements:
well @mcollina that depends on how you interpret the use of await, if you think of it as enclosing all code after the await in an `after` cb then yes it is, but if you think of it enclosing only that which is awaited in `after` then its more like 2

delvedor:
> Ahum, I think option 1 is better and it's compatible to how .after() wroks right now.

Yes, it's similar on when you will access the decorators, but the moment when the plugin is executed is not.

mcollina:
> Yes, it's similar on when you will access the decorators, but the moment when the plugin is executed is not.

in 1), the `first` plugin is run together with `second` when we have an after/await. In 2), the `first` plugin is run when `.ready()` is called. However `first` was defined _before_ `second`. I'm not convinced that 2) is intuitive.

davidmarkclements:
In the context of replicating `after` 1 may be more intuitive, but in the context of thinking about a queue + `await`:
```js
queueAThing()
await aThing()
queueAThing()
await processTheQueue()
```
what's intuitive in this case? that both `queueAThing` ops run on `await processTheQueue` or that the first `queueAThing` op runs before `aThing` because there's an `await`? (it's the first right..)

mcollina:
The problem is that `await` is external. consider this case:

```js
app.use(first)
const p = app.use(second)
app.use(third)
await p // which one will be loaded?
await app.ready()
```

Note that the order in which plugins are loaded in scenario 2 depends on when `.then()` is called. I prefer a more deterministic order based on the order of calls to `.use()`

davidmarkclements:
In your example under 2, it'll be `second`, then `first`, then `third`. Under 1 it'll be first second and third, but I think that's unintuitive. If you assign to `p` and await `p` somewhere else you emphatically want `p` (`second`) to load before anything else.

davidmarkclements:
@mcollina @delvedor I've added tests (and some bug fixes) - behaviour is as per 1, tests show behaviour is as per 1.

If we decide on 1 this is probably ready to go, if we decide on 2 (which I'm actually more in favour of) then I'll adapt it for that. My vote is for 2, if you @delvedor @jsumners and @mcollina can come to a consensus I'll act accordingly.

jsumners:
I think it's fair to say that if you do not `await aThing()` then you do not have any guarantee `aThing()` will be processed before doing `anotherThing()`. Nor should `await anotherThing()` have any dependency on `aThing()` completing before it can return with its result. Which is to say, option 2 makes the most sense.

mcollina:
> Can you also run a benchmark?

What would you benchmark? This will sit in the hot path only for the serverless case, and we have a massive bottleneck in Ajv already for Fastify.

> I’m think that if we go for option 2, then users will start awaiting everything...

I agree, that's one of the reasons I prefer 1.

jsumners:
My opinion is not strongly held. You guys know this code better than I do.

allevo:
I'm not convinced. Maybe I prefer something like
```js
app.use(first)
// app.firstDecorator = undefined
app.use(second)
// app.firstDecorator = undefined
// app.secondDecorator = undefined
app.use(third)
// app.firstDecorator = undefined
// app.secondDecorator = undefined
// app.thirdDecorator = undefined

await app.loadAllButIWantToDefineOtherThings()
// app.firstDecorator = true
// app.secondDecorator = true
// app.thirdDecorator = true
```

Other consideration:
- till now `ready` was emitted once. This is a huge breaking change.
- `use` and `register` return a promise which mean I can catch it. I don't like it
- the load order should be clear

My 2 cents: a method to that explicitly.

davidmarkclements:
Thanks @allevo - we could do `await after` instead... 

```js
app.use(first)
app.use(second)
app.use(third)
await app.after()
// do more things
```

* I'm not seeing where 'ready' is emitted? there's `preReady` and that'll only be emitted once (as far as I understand it), sorry if I'm not seeing something obvious, could you expand on this?
* the catch won't work because the error is propagated to the outer plugin and caught the normal way (as with any error in a child plugin)
* whichever way we do it, the load order will be clear to some and not to others because bias is playing a major role in what feels "intuitive" in this case. The only way to make it clear is to pick an option and document it

davidmarkclements:
There are now 2 options, with one of them having two sub-options

1. `await` is similar to `after`
  a. `await use` - when a plugin is awaited the queue is drained, and proceeds after plugins are loaded
  b. `await after` - explicitly say: I want all the plugins to be loaded to this awaited point
2. `await use` a single plugin (and it's children), treat any non-awaited plugins as additions to the queue

mcollina:
I think a good starting point is to have `await app.after()` work as defined in option 1.

```js
app.use(first)
// app.firstDecorator = undefined

 app.use(second)
// app.firstDecorator = undefined
// app.secondDecorator = undefined

await app.after()
// app.firstDecorator = true
// app.secondDecorator = true

app.use(third)
// app.thirdDecorator = undefined

await app.ready()
// app.firstDecorator = true
// app.secondDecorator = true
// app.thirdDecorator = true
```

We can now debate a second PR to make `await app.use()` a thing.

delvedor:
I don't like `await app.after()`, because while it solves the debate "what should I await" it forces the user to use another API for something that could work without the need to use it.

`.after` has always been a pain point for the Fastify API, because the only people who know what is needed for, it's who has developed the code!
Every time @mcollina and I did a workshop or a talk, very few people understood when to use `after` or why it was there at all.
I think it is time that we remove this complexity because while with callbacks it is necessary, with promises is just code for the sake of code.

mcollina:
I think we need `await app.after()` or `await app.loadPluginsSoFar()` as a low-level mechanism. This includes a bit too much hackyness with the proto to be a solution that can be used 100% of the times (I suspect there will be a 1% situation where this is unfeasible).

IMHO we need both.

davidmarkclements:
I just thought intensely about not using proto, I can't think of a way because you have to be able to link the plugin to the `then` when it's called. With proto and closure this is easy, but if you tried to say, well `server` must have `then` method and a way to push plugins that or may not resolve, then the tracking of which plugin an await is intending to resolve would be impossible given all the edge cases. 

So yeah, I'm inclined to agree with @mcollina - both.

davidmarkclements:
Although, we can't make this work with `after` without proto while also keeping it chainable.... so you have the same problem. So you can either

* break after - make it promise returning and no longer chainable
* introduce a new API that is promise returning

mcollina:
I'm happy to break after chainability.

davidmarkclements:
@mcollina - only saw your comment after pushing..

Well.. I added this one called `assimilate` - this way it's non breaking and we don't have to wait for next fastify major to integrate

```
app.use(plugin1)
```

Ethan-Arrowood:
I have no strong opinions about this. I do like the idea of being able to `await` some set of plugins to be loaded before loading other ones. But I do not know enough about this to weigh in heavily on either option.

davidmarkclements:
ok updated again. So `after(func)` is still chainable, but `after()` returns a promise which resolves to the instance. So it won't break for anyone using `after` the usual way. Also `(await app.after()).use(plugin)` should also work because it'll async resolve back to `app`.

davidmarkclements:
This is actually now really easy to message: 

* `await after()` is the promise based analogue to `after(func)`
* `await use()` is short hand to `app.use(); await after()`

That's as clear as you're going to get I think.

boot.js
```diff
@@ -187,11 +195,19 @@ function assertPlugin (plugin) {
...
 }
 
+Boot.prototype.assimilate = function (plugin) {
+  plugin = plugin || this._lastUsed
+  return new Promise((resolve, reject) => {
+    if (plugin) {
+      plugin.asyncQ.push(() => resolve(this._server))
+    } else resolve(this._server)
+  })
+}
```
mcollina: This is still called assimilate

boot.js
```diff
@@ -27,14 +27,22 @@ function wrap (server, opts, instance) {
...
+    instance.use(fn, opts)
+    const plugin = instance._lastUsed
+    const thenableDescriptor = {
+      then: {
+        value (resolve) {
+          return instance.assimilate(plugin).then(resolve)
+        }
+      }
+    }
+    return Object.create(this, thenableDescriptor)
```
mcollina: I think this is not documented right now.

boot.js
```diff
@@ -27,14 +27,22 @@ function wrap (server, opts, instance) {
...
-  server[useKey] = function (a, b, c) {
-    instance.use(a, b, c)
-    return this
+  server[useKey] = function (fn, opts) {
+    instance.use(fn, opts)
+    const plugin = instance._lastUsed
+    const thenableDescriptor = {
+      then: {
+        value (resolve) {
+          return instance.assimilate(plugin).then(resolve)
```
mcollina: let's not call it assimilate.

davidmarkclements: ahahahahahah

boot.js
davidmarkclements: https://github.com/mcollina/avvio/pull/92/files#diff-04c6e90faac2675aa89e2176d2eec7d8R148

davidmarkclements: https://github.com/mcollina/avvio/pull/92/files#diff-04c6e90faac2675aa89e2176d2eec7d8R167

davidmarkclements:
I just checked this out, it's now passing on Node 6 and Node 8 - switching back to master because this could ultimately be fully compat with fastify v2

boot.js
davidmarkclements: "thenanble wrapped instance" should suffice?

davidmarkclements:
@mcollina done - but will this still land in fastify v2 or we looking at v3 regardless of node version compat?

mcollina:
I'd be very careful in landing this in v2 of Fastify so late in the maintainance game. I prefer to land it in v3.

mcollina:
I would prefer that as well before landing.

README.md
```diff
@@ -292,6 +340,38 @@ app.after(async function () {
 
 Returns the instance on which `after` is called, to support a chainable API.
 
+-------------------------------------------------------
+<a name="await-after"></a>
+
+### await app.after() | app.after() => Promise
+
+Calling after with no function argument loads any plugins previously registered via `use` and returns a promise which resolves when all plugins registered so far have loaded.
```
Eomm: Should we specify that `await after()` will never reject?

mcollina: I actually _missed_ that this would never reject. I'm not really convinced of this behavior.
Should it reject instead?

Ethan-Arrowood:
> Can we open a pr in Fastify where we use this branch so we can be sure that we are not breaking existing code?

I'll try to do this right now

Ethan-Arrowood:
Currently breaks one test from `decorator.test.js`:

https://github.com/fastify/fastify/blob/next/test/decorator.test.js#L665-L712

davidmarkclements:
thanks @Ethan-Arrowood for doing that! 

ok I'm confused by this test, it's my understanding that `after` should run after all registers in the current tick. 

In the failing test we're doing an after, then a decorate. Then as a sibling to the after, we're doing a register and expecting the decorator to be present? I'm not trying to be arrogant but is it possible that this test is incorrect and there is a bug in current? Or am I failing to understand how after is supposed to work?

mcollina:
> ok I'm confused by this test, it's my understanding that after should run after all registers in the current tick

That is not correct with the behavior in master. Each function passed to after is queued in the same queue of the plugins, and as a result it should be executed _before_ the second register.

davidmarkclements:
gotcha, okay

boot.js
```diff
@@ -27,14 +27,19 @@ function wrap (server, opts, instance) {
     throw new Error(readyKey + '() is already defined, specify an expose option')
   }
 
-  server[useKey] = function (a, b, c) {
-    instance.use(a, b, c)
-    return this
+  server[useKey] = function (fn, opts) {
+    instance.use(fn, opts)
+    const plugin = instance._lastUsed
+    instance.then = (resolve) => {
```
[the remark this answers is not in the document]
davidmarkclements: yeah and a `catch` method. Promise returned from `_loadRegistered` has no rejection in its still currently a mystery to me on how to determine the failure mode of this case. A throw in a plugin that has been await registered does not end up on the `_error` property. Any tips appreciated

mcollina:
@delvedor @Eomm @jsumners would you mind doing a final check before I land?

mcollina:
I think we should start.

Il giorno mar 11 feb 2020 alle ore 22:37 Manuel Spigolon <
notifications@github.com> ha scritto:

Eomm:
Oh, and I forgot to tell that, in the example, if you replace ready with start and autoload:true, an error is thrown

mcollina:
I'm working on a fix for the problems you flagged @Eomm and one more that we have if `override` is used. I'll send a PR to Dave's branch.

mcollina:
@Eomm this should now work as expected, thanks for the review.

boot.js
```diff
@@ -187,11 +203,45 @@ function assertPlugin (plugin) {
...
+  this._lastUsed = this._addPlugin(plugin, opts, false)
   return this
 }
 
+Boot.prototype._loadRegistered = function (plugin) {
+  plugin = plugin || this._lastUsed
+  return new Promise((resolve) => {
+    if (plugin && !plugin.loaded) {
+      plugin.asyncQ.push((s) => {
+        // TODO check for override support
```
davidmarkclements: there's a TODO here

boot.js
```diff
@@ -187,11 +203,45 @@ function assertPlugin (plugin) {
...
+
+Object.defineProperty(Boot.prototype, 'then', {
+  get () {
+    // If the instance is ready, then there is
+    // nothing to await. This is true during
+    // await server.ready() as ready() resolves
+    // with the server, end we will end up here
+    // because of automatic promise chaining.
+    if (this.booted) {
+      return undefined
```
davidmarkclements: we don't need to explicitly return `undefined` - also is there a way to reuse this for server.then?

boot.js
davidmarkclements: addressed - was already done

boot.js
davidmarkclements: done, see `thenify` function

davidmarkclements:
ok I've had one last sweep and addressed nits. LGTM to me from here - cc @mcollina

PR #95, opened 2020-02-18
Await fixes

### LINKED ISSUE #93: Do not load any more plugins with .after() when the previous errors
From Fastify https://github.com/fastify/fastify/pull/2093/files:

```js
test('awaitable after error handling', async t => {
   const fastify = Fastify()
   let first = false
   let second = false
   let third = false

   // Note that this does not throw in case of error,
   // all errors are currently collected to ready() because it's
   // not really possible to do anything about them.
   await fastify.register(async (instance, opts, next) => {
     first = true
     throw new Error('kaboom')
   })
   t.is(first, true)

   fastify.register(async (instance, opts, next) => {
     second = true
   })

   await fastify.after()
   // TODO this is a bug, it should be false
   t.is(second, true)

   fastify.register(async (instance, opts, next) => {
     third = true
   })

   await t.rejects(fastify.ready())
   t.is(third, false)
 })
```

cc @davidmarkclements

### LINKED ISSUE #94: Return avvio inside after breaks avvio
If you use `avvio#next`, there is a rather funny bug, if you return avvio inside an after function without callback (I guess it will happens the same inside a use), avvio will go crazy and stop its execution.

```js
'use strict'

const avvio = require('.')()

avvio.use((instance, opts, next) => {
  console.log('plugin 1')
  next()
})

avvio.after(() => {
  console.log('after plugin 1')
  return avvio // <=================== LOOK HERE
})

avvio.use((instance, opts, next) => {
  console.log('plugin 2')
  next()
})

avvio.ready(err => {
  console.log(err || 'ready')
})

/*

Actual log:

plugin 1
after plugin 1



Expected log:

plugin 1
after plugin 1
plugin 2
ready

*/
```

This is very likely caused by how we are handling promises and the fact that avvio itself is a thenable.
https://github.com/mcollina/avvio/blob/5d36884e9f2c4483a23466668a41a4edb9e9ff4c/boot.js#L434-L451

cc @davidmarkclements

mcollina:
This PR changes substantially how our `await app.register()` and `await app.after()` works. Essentially, it's impossible to "skip error and continue", as we were doing before. Instead, this makes the promise reject.

Fixes #93 
Fixes #94

### DIFF
--- a/test/await-after.test.js
+++ b/test/await-after.test.js
@@ -27,6 +27,27 @@ test('await after - nested plugins with same tick callbacks', async (t) => {
   t.pass('reachable')
 })
 
+test('await after without server', async (t) => {
+  const app = boot()
+
+  let secondLoaded = false
+
+  app.use(async (app) => {
+    t.pass('plugin init')
+    app.use(async () => {
+      t.pass('plugin2 init')
+      await sleep(1)
+      secondLoaded = true
+    })
+  })
+  await app.after()
+  t.pass('reachable')
+  t.is(secondLoaded, true)
+
+  await app.ready()
+  t.pass('reachable')
+})
+
 test('await after - nested plugins with future tick callbacks', async (t) => {
   const app = {}
   boot(app)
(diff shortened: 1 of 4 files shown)


mcollina:
cc @davidmarkclements

test/await-after.test.js
```diff
@@ -202,7 +223,7 @@ test('await after - error handling, future tick cb err, nested', async (t) => {
   t.rejects(() => app.ready(), Error('kaboom'))
 })
 
-test('await after complex scenario', async (t) => {
+test('await after complex scenario', { only: true }, async (t) => {
```
Eomm: Is this wanted?

PR #96, opened 2020-02-20
add preready managing errors

Eomm:
Ref to:
https://github.com/fastify/fastify/pull/2077


Eomm:
Do you think this PR could land to go further with the fastify's issue?
Thks

PR #97, opened 2020-02-25
Actually rejects when await app.after() or await app.register()

mcollina:
In #95, I _thought_ I did this. Turns out I was missing a critical change, this is it.

### DIFF
--- a/test/await-after.test.js
+++ b/test/await-after.test.js
@@ -134,55 +134,59 @@ test('await after - promise returning function plugins + promise chaining', asyn
   t.pass('reachable')
 })
 
-test('await after - error handling, async throw', async (t) => {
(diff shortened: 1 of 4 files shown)


mcollina:
cc @Eomm please check

mcollina:
This is needed to solve https://github.com/fastify/fastify/pull/2114.

test/await-after.test.js
```diff
@@ -134,93 +134,99 @@ test('await after - promise returning function plugins + promise chaining', asyn
   t.pass('reachable')
 })
 
-test('await after - error handling, async throw', async (t) => {
+test('await after - error handling, async throw', { only: true }, async (t) => {
```
Eomm:
```suggestion
test('await after - error handling, async throw', async (t) => {
```

PR #100, opened 2020-03-15
No async q - use a Promise instead

mcollina:
This PR simplifies the asyncq logic.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -205,27 +205,19 @@ Boot.prototype.use = function (plugin, opts) {
 
 Boot.prototype._loadRegistered = function () {
   const plugin = this._current[0]
-  return new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


boot.js
```diff
@@ -205,27 +205,19 @@ Boot.prototype.use = function (plugin, opts) {
...
-    }
-  })
+  // if the root plugin is not loaded, let's resume that
+  // so one can use after() befor calling ready
+  if (weNeedToStart) {
+    process.nextTick(() => this._root.q.resume())
+  }
+
+  if (!plugin) {
+    return Promise.resolve()
```
davidmarkclements: does this happen a lot - if so is it worth re using a dummy promise?

mcollina: It's a pretty rare condition.

PR #102, opened 2020-03-29
After override

### LINKED ISSUE #101: Are calls to override wanted?
Reading the docs (master&next) of the `after`function:

> If three parameters are given to the callback, the first will be the error object, the second will be the top level context unless you have specified both server and override, __in that case the context will be what the override returns__, and the third the done callback.

My understanding is that the `.after` interfaces with one or two params don't call the `override`
Instead `override` is always called:

```js
'use strict'

const server = { count: 0 }
const avvio = require('.')(server)

avvio.override = function (s, fn, opts) {
  console.log('override', s.count)
  const res = Object.create(s)
  res.count = res.count + 1
  return res
}

avvio
  .use(function hello (server, opts, done) {
    console.log('hello server', server.count)
    done()
  })
  .after(function (err, done) {
    if (err) throw err
    done()
  })
  .after(function (err, context, done) {
    if (err) throw err
    console.log('context', context.count)
    done()
  })

avvio.ready(() => { console.log('ready') })
```

Show:

```
override 0
hello server 1
override 0
override 0
context 0
ready
start
```

is this wanted/needed?
Does not this behaviour add unuseful overhead?

With this (fast) code on `plugin.js` all our tests are green and also on fastify.next all is green:

```
if (!this.isAfter || func.length === 3 || (func.length === 2 && func.then)) {
  this.server = this.parent.override(server, func, this.opts)
}
```

Doubt born by https://github.com/fastify/fastify/pull/2166

Eomm:
I lied 😅
My snippet didn't check the all the after's functions were wrapped in a function that had always 3 params.

Now the flow that calls `.after()` without any parameter doesn't work.. still working on it

Fixes #101

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -142,10 +142,7 @@ Plugin.prototype.loadedSoFar = function () {
   }
 
   const setup = () => {
-    console.log('.............', this.server)
     this.server.after((err, cb) => {
-      console.log('.............')
-
       this._error = err
       this.q.pause()
       debug('resolving promise', this.name)
--- a/plugin.js
+++ b/plugin.js
@@ -167,6 +167,10 @@ Plugin.prototype.loadedSoFar = function () {
 
     if (!this.server) {
       this.on('start', setup)
+      this.parent.on('error', (err) => {
+        this._promise.reject(err)
+        this._promise = null
+      })
     } else {
       setup()
     }
@@ -199,6 +203,7 @@ Plugin.prototype.finish = function (err, cb) {
     if (this._promise) {
       this._promise.reject(err)
       this._promise = null
+      this.emit('error', err)
     }
     done()
     return
@@ -212,7 +217,13 @@ Plugin.prototype.finish = function (err, cb) {
           debug('wrap')
           queueMicrotask(check)
         }
-        this._promise.resolve()
+
+        if (this._error || (this.parent && this.parent._error)) {
+          this._promise.reject(this._error || (this.parent && this.parent._error))
+        } else {
+          this._promise.resolve()
+        }
+
         this._promise.promise.then(wrap, wrap)
         this._promise = null
       } else {
--- a/plugin.js
+++ b/plugin.js
@@ -73,13 +73,14 @@ Plugin.prototype.exec = function (server, cb) {
     return
   }
 
-  try {
-    if (!this.isAfter || this.isAfter === 3) {
+  if (!this.isAfter) {
+    // Skip override for after
+    try {
       this.server = this.parent.override(server, func, this.opts)
+    } catch (err) {
+      debug('override errored', name)
+      return cb(err)
     }
-  } catch (err) {
-    debug('override errored', name)
-    return cb(err)
   }
 
   this.opts = typeof this.opts === 'function' ? this.opts(this.server) : this.opts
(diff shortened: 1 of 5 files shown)


Eomm:
Right now I didn't found a solution

The concept is that when we use `after()` or `await .use` (it is the same):
- `_loadRegistered` start loading
- if there is not a `server` (like our case), an event is registered:

https://github.com/mcollina/avvio/blob/4011f5a955352b247acf086bdd0ca487bb0d646b/plugin.js#L167

- this event is never executed nor, if it would, it would break since `this.server` is not being created

For this reason, I was trying to listen to the error event of the parent server that is being created, but also these change doesn't cover all the tests 😕

Eomm:
> would it simplify things to avoid override in after completely? I would actually prefer that.

Do you mean to avoid calling `_after` internally?

plugin.js
```diff
@@ -74,7 +74,9 @@ Plugin.prototype.exec = function (server, cb) {
   }
 
   try {
-    this.server = this.parent.override(server, func, this.opts)
+    if (!this.isAfter || this.isAfter === 3) {
```
mcollina: I think we should just check if isAfter is true.

Eomm: The failing errors are the same

mcollina:
I have a feeling we should never trigger override with after.

Eomm:
Ok, I'm struggling with 2 tests and then I will clean up the code removing the override call for after

mcollina:
should I give it a go?

Eomm:
It would be helpful 👍 

The failing test is `await after complex scenario`:

> root plugin has already booted

The issue is that I can't distinguish from an `await after` `await use` from an `await ready`.

```
await app.after() // this set the application as booted
...
app.use(second) // this throws the error
```

I was working on `plugin.loadedSoFar`: this diff fixes the "already booted" problem, but doesn't load all the plugins, I didn't find how I should pause/resume the queue.

```diff
@@ -128,7 +129,7 @@ Plugin.prototype.exec = function (server, cb) {
   var promise = func(this.server, this.opts, done)
 
   if (promise && typeof promise.then === 'function') {
-    debug('resolving promise', name)
+    debug('resolving promise A', name)
 
     promise.then(
       () => process.nextTick(done),
@@ -145,7 +146,7 @@ Plugin.prototype.loadedSoFar = function () {
     this.server.after((err, cb) => {
       this._error = err
       this.q.pause()
-      debug('resolving promise', this.name)
+      debug('resolving promise B', this.name)
 
       if (err) {
         this._promise.reject(err)
@@ -166,17 +167,33 @@ Plugin.prototype.loadedSoFar = function () {
     res = this._promise.promise
 
     if (!this.server) {
-      this.parent.once('finish', (err) => {
-        this._error = err
-        this.q.pause()
-        debug('resolving promise', this.name)
-        if (err) {
-          this._promise.reject(err)
-        } else {
-          this._promise.resolve()
-        }
-        this._promise = null
-      })
+      let i = 0
+      this.q.resume()
+      if (!this.parent.alredyRegistered) {
+
+        this.parent.alredyRegistered = true
+        this.parent.on('finish', (err) => {
+          this.q.pause()
+          if (i === 0 && !err) { // the first is "bound root" itself
+            i++
+            return
+          }
+
+          this._error = err
+          if (!this._promise) {
+            return
+          }
+          debug('resolving promise C', this.name)
+          if (err) {
+            this._promise.reject(err)
+          } else {
+            this._promise.resolve()
+          }
+          this._promise = null
+        })
+      }
     } else {
       setup()
     }
```

mcollina:
Can you add me to your repo so I can push there?

mcollina:
@Eomm done, I think this should work. There was an "old" bug that made `isAfter` becoming `function noop () {}` for root.

Eomm:
Great!

Let me update the docs and then this PR will be ready!

PR #104, opened 2020-04-11
fix timeout on ready

### LINKED ISSUE #103: Ready ignore timeout
The `ready` queue doesn't apply the `timeout` parameter.

Is this wanted?

```
const boot = require('..')

const app = boot({ timeout: 1000 })

// ready with two parameter
app.ready(function (_, done) {
  console.log('ready called')
  setInterval(() => {}, 100)
  // done() // not call done
})

app.on('start', () => {
  // never called
  console.log('start')
})

app.start()
```

Eomm:
Fixes #103 

This should be backported to master as well

Ref https://github.com/fastify/fastify/pull/2166

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -386,10 +386,47 @@ function callWithCbOrNextTick (func, cb, context) {
     } else {
       process.nextTick(cb)
     }
-  } else if (func.length === 2) {
-    func(err, cb)
   } else {
-    func(err, context, cb)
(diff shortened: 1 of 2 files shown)


mcollina:
Can you target master instead? It'd be easier to forward-port right now.

mcollina:
I've landed the change for master and merged that in next, and released that as well.

PR #106, opened 2020-04-12
Allow after to swallow/handle errors

mcollina:
Fixes: https://github.com/fastify/fastify/issues/2194

### DIFF
--- a/test/after-and-ready.test.js
+++ b/test/after-and-ready.test.js
@@ -753,3 +753,27 @@ test('preReady event (errored)', (t) => {
     t.is(order.shift(), 3)
   })
 })
+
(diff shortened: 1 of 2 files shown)


test/after-and-ready.test.js
```diff
@@ -753,3 +753,27 @@ test('preReady event (errored)', (t) => {
     t.is(order.shift(), 3)
   })
 })
+
+test('after 1 param swallows errors with server and timeout', { only: true }, (t) => {
```
Eomm, line 757:
```suggestion
test('after 1 param swallows errors with server and timeout', (t) => {
```

PR #126, opened 2021-02-10
fix typescript declaration for app.use with async functions

tellnes:
#### Checklist

   - `npm run benchmark` fails with `missing script: benchmark`
     - not relevant
- [x] documentation is changed or added
    - not relevant

Btw, the test fails with an `listen EADDRINUSE: address already in use :::3000` if there is something else listening on port `3000` where the tests is run.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,7 +25,7 @@ declare namespace avvio {
   }
 
   interface Plugin<O, I> {
-    (server: context<I>, options: O, done: (err?: Error) => void): void;
+    (server: context<I>, options: O, done: (err?: Error) => void): unknown;
   }
 
   interface Server<I> {
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -369,3 +369,10 @@ import * as avvio from "../../";
   // avvio with server, options and done callback
   const app = avvio(server, options, () => undefined);
 }
+
+{
+  const app = avvio();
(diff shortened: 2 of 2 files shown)


climba03003:
I think it really doesn't matter to the return value to be `void` or `Promise<void>`
It is valid for `void` itself only.
image

tellnes:
It does matter if you are running linting with the [@typescript-eslint/no-misused-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md) rule enabled.

But I think it should be `unknown` instead of `void`. That will allow you to use promises which returns something also. That will be fine since avvio seems to ignores any return value. I will update the PR.

tellnes:
PR updated. Changed return value from `void | Promise<void>` to `unknown | Promise<unknown>`

mcollina (changes requested):
Would you mind to add a test in https://github.com/fastify/avvio/tree/master/test/types for this?

tellnes:
@mcollina Since this is to fix a linting rule, I don't think I will be able to add a failing test case without configuring eslint for this project.

tellnes:
Test case added. I did also remove `Promise<unknown>`. That is not needed anymore after I also changed the non-async part from `void` to `unknown` in an earlier change. The change is now just to change it from `void` to `unknown`.

mcollina (approved):
lgtm

PR #140, opened 2021-04-27
Improve plugin getName() speed

airhorns:
While profiling some slow tests and slow boot times for my application, I noticed `avvio`'s plugin naming stuff at the top of the profile for the first test case executed. I have a lot of fastify plugins (say 500 or so), and I think getting the name for all of them starts to get a bit slow because it is O(n^2). This switches the plugin naming to use a prebuilt cache of names with a WeakMap instead of iterating require.cache every time.

Here's the benchmark I used:

```javascript
const Benchmark = require('benchmark')
const plugin = require('./plugin')
const glob = require('glob')
const modules = glob.sync('./node_modules/*/')

console.log(`Benchmarking against ${modules.length} modules`)
const suite = new Benchmark.Suite()

suite.add('getName()', function () {
  for (const module of modules) {
    plugin.getName(module)
  }
})
  .on('cycle', function (event) {
    console.log(String(event.target))
  })
  .run()
```

On `master`:

```
inspector ~/C/avvio (master*) ➜  node name-bench.js
Benchmarking against 392 modules
getName() x 1,818 ops/sec ±1.68% (90 runs sampled)
```

On `faster-plugin-naming`:

```
inspector ~/C/avvio (faster-plugin-naming*) ➜  node name-bench.js
Benchmarking against 392 modules
getName() x 8,217 ops/sec ±0.51% (89 runs sampled)
```

Note if you want to run the benchmark yourself you'll have to `npm install benchmark` and export the `getName` function from the `plugin` module. 

I am not aware of a way to do this without using the deprecated `require.extensions` property in node, but, iterating the require cache every time a plugin is named seems excessive so it is justified to me. Happy to hear your feedback though.

#### Checklist

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -7,17 +7,26 @@ const inherits = require('util').inherits
 const debug = require('debug')('avvio')
 const CODE_PLUGIN_TIMEOUT = 'ERR_AVVIO_PLUGIN_TIMEOUT'
 
-function getName (func) {
-  // let's see if this is a file, and in that case use that
-  // this is common for plugins
-  const cache = require.cache
-  const keys = Object.keys(cache)
-
-  // eslint-disable-next-line no-var
-  for (var i = 0; i < keys.length; i++) {
-    if (cache[keys[i]].exports === func) {
-      return keys[i]
+/* eslint-disable node/no-deprecated-api */
+const modules = new WeakMap()
+const superLoaders = { ...require.extensions }
+for (const extension of Object.keys(require.extensions)) {
+  require.extensions[extension] = function (module, filename) {
+    const result = superLoaders[extension](module, filename)
+    if (module.exports && typeof module.exports === 'function') {
+      modules.set(module.exports, filename)
     }
+    return result
+  }
+}
+/* eslint-enable node/no-deprecated-api */
+
+function getName (func) {
+  // let's see if this is function is the exports of a file, and in that case use that file's name
+  // this is common for plugins
+  const modulePath = modules.get(func)
+  if (modulePath) {
+    return modulePath
   }
 
   // if not maybe it's a named function, so use that


mcollina:
Are you loading your plugins with `fastify-autoload`? In case we should make sure you never hit this method.

I'm not fond of the current code and the fact that it would only work for commonjs and not esm. I think we should think of  refactoring this by drastically changing the algorithm. _I think we should eventually drop it_.

A slightly worse solution that do not use the deprecated API is computing a `WeakMap` of all the modules on first call of `getName()` (and then every time there is a miss).

airhorns:
> Are you loading your plugins with fastify-autoload?

No, I'm using my own loader thing because I needed more control over load order and encapsulation than fastify-autoload had available

> I'm not fond of the current code and the fact that it would only work for commonjs and not esm. I think we should think of refactoring this by drastically changing the algorithm. I think we should eventually drop it.

Same, not ideal. We could use just `function.name` if it exists and the function text otherwise?

> A slightly worse solution that do not use the deprecated API is computing a WeakMap of all the modules on first call of getName() (and then every time there is a miss).

That would work, but I think there will still be a lot of misses because by definition this code is getting run during code loading, which means the require cache is not going to be full yet.

mcollina:
You should do https://github.com/fastify/fastify-autoload/blob/d0d93a542c38aba4ce281bf3e61304d2cc8efed0/index.js#L252. I think fastify-autoload manage this automatically without triggering the problem.

airhorns:
I can do that, but, other plugins aren't always going to register themselves with a name, do we really want to leave this O(N^2) traversal in here?

mcollina:
This is there as last-measure scenario. The only way to provide a good user experience is with the auto-loading libraries setting those metadata properly.

I'm happy to remove it, but we should update the docs somewhere in FAstify clarifying that they should specify the name somehow.

airhorns:
Ok, I will close and open a docs PR at some point.

airhorns:
This came up again in another profiling session for me locally and I noticed something I missed before. As far as I can tell, we always call the `getName` function on any functions passed to avvio plugins: https://github.com/fastify/avvio/blob/master/plugin.js#L50 . So, even if a fastify plugin specifies a name in the fastify plugin opts, avvio still figures out its own name in the constructor using this slow function. Is there a `opts.name || getName(func)` bit missing? Or am I missing something in one of the downstream packages?

mcollina:
That would be a good optimization! Could you check it this would work?

PR #162, opened 2021-11-05
Use fastify plugin metadata or an explicit passed in name option for a plugin's name

airhorns:
This is a second approach to the same thing in #140.

`getName` in avvio gets a string name for each plugin registered to use when reporting debug info. Doing some profiling of a big fastify application, I noticed that a lot of time was spent getting the name of avvio plugins in the `getName` function. For me, `require.cache` was very big, and the number of registered plugins was very big, so the O(n^2) search for each file in the require cache was actually taking a good long while. Also, `getName` relying on the require cache is undesirable as ESM becomes more and more common -- we should try to move to an approach that doesn't need that.

Fortunately, a lot of plugins registered with avvio already report their names! We were somewhat accidentally using `getName` to get the name of a lot of plugins that already have fastify metadata set, or pass an explicit name in with their options. We can avoid the nasty performance pitfall by just accepting the incoming names inside avvio, and then falling back on one of the much faster last-resort strategies if plugin names aren't set.

It is kind of unfortunate that we can't use the plugin's file path as a nice, human friendly name for a plugin that isn't otherwise named, but I think the performance issue merits its removal, and I will add documentation to fastify to indicate that naming your own plugins is a good idea.

I tried running the fastify test suite with this version of this library in place, and the only things that failed were like this:

```
  --- expected
  +++ actual
  @@ -1,1 +1,1 @@
  -ERR_AVVIO_READY_TIMEOUT
  +AVV_ERR_READY_TIMEOUT
```

which I think is from other changes to `avvio` master, not this change. 

#### Checklist

### DIFF
--- a/test/fixtures/plugin-no-next.js
+++ b/test/fixtures/plugin-no-next.js
@@ -1,5 +1,5 @@
 'use strict'
 
-module.exports = function (app, opts, next) {
+module.exports = function noNext (app, opts, next) {
   // no call to next
 }
--- a/plugin.js
+++ b/plugin.js
@@ -7,20 +7,20 @@ const inherits = require('util').inherits
 const debug = require('debug')('avvio')
 const { AVV_ERR_READY_TIMEOUT } = require('./lib/errors')
 
-function getName (func) {
-  // let's see if this is a file, and in that case use that
-  // this is common for plugins
-  const cache = require.cache
-  const keys = Object.keys(cache)
+// this symbol is assigned by fastify-plugin
+const kPluginMeta = Symbol.for('plugin-meta')
 
-  // eslint-disable-next-line no-var
-  for (var i = 0; i < keys.length; i++) {
-    if (cache[keys[i]].exports === func) {
-      return keys[i]
-    }
+function getName (func, optsOrFunc) {
+  // use explicit function metadata if set
+  if (func[kPluginMeta] && func[kPluginMeta].name) {
+    return func[kPluginMeta].name
   }
 
-  // if not maybe it's a named function, so use that
+  if (typeof optsOrFunc !== 'undefined' && typeof optsOrFunc !== 'function' && optsOrFunc.name) {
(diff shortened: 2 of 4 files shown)


test/fixtures/plugin-no-next.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
 
-module.exports = function (app, opts, next) {
+module.exports = function noNext (app, opts, next) {
```
jsumners, line 3: This is a key change. We should still support assigning a fallback name to anonymous functions. I do believe this is required to facilitate showing which plugin throws an error in Fastify.

airhorns: This change definitely still assigns a fallback name, just using a different strategy than before. Previously it'd use the name of the file the function was defined in if the function is any file's default export, and otherwise it'd use the name of the function if it is named, and as a last resort it'd use the first two lines of the functions source. Now, we use metadata as the first strategy, and then the name if it exists, and then the source. This diff is just to make the test a bit terser.

jsumners: But this change loses the test covering anonymous functions.

mcollina: @airhorns could you add a new test?

mcollina:
This targets the new avvio release that will be part of Fastify@v4, don't worry about the broken test.

test/fixtures/plugin-no-next.js
airhorns: Ok, updated to add a test, I still think this file should be changed though as it pertains to timeout tests, not plugin naming

mcollina (approved):
lgtm

Eomm (approved):

mcollina (approved):
lgtm

mcollina:
cc @jsumners you ok now?

jsumners (approved):

PR #173, opened 2022-02-24
fix: delay plugin promise resolution until next tick

chrskrchr:
#### Checklist

#### Description
This PR proposes a fix for the plugin registration race condition we're seeing in this `fastify` issue :

https://github.com/fastify/fastify/issues/3726

The TL;DR; is that it appears the use of a top-level `await` between some (but not all) plugin registrations can get `avvio`'s root plugin into a weird state where it resolves an `after()` promise prior to in-flight registration functions having completed.  The issue doesn't occur if a top-level `await` statement exists between either ALL plugins or NONE of them.

I compared the `DEBUG=avvio` output between a "bad" boot sequence (e.g., what's in the [sample repo](https://github.com/chrskrchr/fastify-plugin-race-condition))  and "good" boot sequence (top-level `await` statements between all plugins) and it appears the issue in the "bad" boot sequence stems from plugin0's `_after` promise being resolved after the plugin1 has started initializing:
 
Screen Shot 2022-02-24 at 12 31 01 PM

This PR proposes a fix where a plugin's promise resolution is always delayed until `process.nextTick`, which appears to give the plugin's `_after` promise a chance to resolve prior to execution moving on to the next plugin.

### DIFF
--- a/test/await-after.test.js
+++ b/test/await-after.test.js
@@ -4,7 +4,7 @@ const { test } = require('tap')
 const boot = require('..')
 const { promisify } = require('util')
 const sleep = promisify(setTimeout)
-const fs = require('fs/promises')
+const fs = require('fs').promises
 const path = require('path')
 
 test('await after - nested plugins with same tick callbacks', async (t) => {
--- a/test/await-after.test.js
+++ b/test/await-after.test.js
@@ -247,7 +247,7 @@ test('await after complex scenario', async (t) => {
   t.notOk(fourthLoaded, 'fourth is not loaded')
   app.use(second)
   t.ok(firstLoaded, 'first is loaded')
-  t.notOk(secondLoaded, 'second is not loaded')
+  t.ok(secondLoaded, 'second is loaded')
   t.notOk(thirdLoaded, 'third is not loaded')
   t.notOk(fourthLoaded, 'fourth is not loaded')
   app.use(third)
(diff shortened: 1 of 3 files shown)


test/await-after.test.js
```diff
@@ -247,7 +247,7 @@ test('await after complex scenario', async (t) => {
   t.notOk(fourthLoaded, 'fourth is not loaded')
   app.use(second)
   t.ok(firstLoaded, 'first is loaded')
-  t.notOk(secondLoaded, 'second is not loaded')
+  t.ok(secondLoaded, 'second is loaded')
```
chrskrchr, line 250: I'm not exactly sure why this behavior changed.  My guess is that delaying the plugin's promise resolution to `process.nextTick` results in a different ordering of items in the root plugin's work queue, allowing the second plugin's function to be called on the same tick as the `app.use(second)` on L250 without having to `await` to allow that function to be called on a future tick.

I defer to the maintainers if this is a concerning change or not.

mcollina: Unfortunately it is :(.

chrskrchr: Is the expectation here that a plugin function is _never_ called on the same tick as when it's registered w/ avvio?  i.e., similar to how the promise spec dictates that a `.then(cb)` callback is to be invoked on the next turn of the event loop, not the current

test/await-after.test.js
```diff
@@ -280,6 +280,54 @@ test('await after complex scenario', async (t) => {
   }
 })
 
+test('without autostart and sync/async plugin mix', async (t) => {
```
Eomm, line 283: Would you mind to add the same test but reading a file using fs instead of the sleep function?

chrskrchr: Do you mean an FS read within a plugin register function?  or in between registrations? (or both? 😄)

chrskrchr: I just pushed a commit that places some of the `sleep` calls with FS reads.  Please let me know if this is/isn't what you had in mind.

test/await-after.test.js
```diff
@@ -4,6 +4,8 @@ const { test } = require('tap')
 const boot = require('..')
 const { promisify } = require('util')
 const sleep = promisify(setTimeout)
+const fs = require('fs').promises
```
chrskrchr, line 7: The most recent CI failure was due to an FS promises import that didn't work with Node 12.  This updated import should work w/ Node 12/14/16

test/await-after.test.js
Eomm: Your update covers all these cases 😄 

Thanks for the check (the timers happens before the IO queue)

Eomm (approved):

Eomm:
Thanks for sharing, we should add more test cases.

The `after` is a key feature used to load the routes in fastify and the validation processing.
Could we run the fastify's test using this branch?

mcollina (approved):
lgtm, this should be ok

I'll backport it to the v7 line

mcollina:
(I run fastify tests too!)

xtx1130 (commented):
LGTM

chrskrchr:
Thanks, everyone, for prioritizing the review and merging of this fix!  This is going to save my organization a bunch of time manually patching our Fastify apps.

@mcollina - I see you already backported this fix to `avvio@3.x`.  Would you like me to submit a PR to `fastify@3.x` with this version bump?  I'll flag the PR as fixing the issue below:

https://github.com/fastify/fastify/issues/3726

climba03003:
> Thanks, everyone, for prioritizing the review and merging of this fix! This is going to save my organization a bunch of time manually patching our Fastify apps.
> 
> @mcollina - I see you already backported this fix to `avvio@3.x`. Would you like me to submit a PR to `fastify@3.x` with this version bump? I'll flag the PR as fixing the issue below:
> 
> [fastify/fastify#3726](https://github.com/fastify/fastify/issues/3726)

I think there is no need for the version bump. The `package.json` do not lock the version and we do not release a particular version for dependency only (unless breaking change). Just reinstall your modules and that should be fine.

chrskrchr:
> I think there is no need for the version bump on fastify@3.x. The package.json do not lock the version and we do not release a particular version for dependency only (unless breaking change). Just reinstall your modules and that should be fine.

@climba03003 - you're absolutely right!  🤦

mcollina:
Unfortunately this broken fastify: https://github.com/fastify/fastify-cors/issues/189.

I'll revert the change on both lines

PR #177, opened 2022-03-01
fix: delay plugin loading until next tick

chrskrchr:
#### Checklist

## Description
This PR is a followup to #173.  That original PR attempted to fix the async plugin loading issue described in [fastify/3726](https://github.com/fastify/fastify/issues/3726) by delaying each plugin's `after` promise resolution until the next tick.  However, this resulted in a subtle but significant change where subsequent plugin registration functions were now being enqueued in `this.q` and executed on the same tick of the event loop as registration, a consequence of the registration function being the only item in `this.q`.  We observed a symptom of this change in the unrelated unit test linked below:

https://github.com/fastify/avvio/pull/173/files#r814200533

We didn't think the above was a major/breaking change in behavior, but that proved to be incorrect per this bug report:

https://github.com/fastify/fastify-cors/issues/189

This PR improves on the initial PR by delaying the entire loading of each plugin to the next tick rather than just the plugin's `after` promise resolution.  This change has been tested in @pvanagtmaal's [test repo]( https://github.com/pvanagtmaal/fastify-test) and produces the expected output for both the working and previously-broken test cases:

```
➜  fastify-test git:(main) ✗ node working.js
cors registered
└── /
    ├── /
    ├── * (OPTIONS)
    └── v
        ├── 1/user (GET)
        └── 2/user (GET)

➜  fastify-test git:(main) ✗ node broken.js
└── /
    ├── /
    ├── * (OPTIONS)
    └── v
        ├── 1/user (GET)
        └── 2/user (GET)
```

It also doesn't require any unexpected changes in [unrelated test cases like the previous PR](https://github.com/fastify/avvio/pull/173/files#r814200533).

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -249,9 +249,7 @@ Plugin.prototype.finish = function (err, cb) {
 // to run prior to executing the next plugin
 function loadPluginNextTick (toLoad, cb) {
   const parent = this
-  process.nextTick(() => {
-    loadPlugin.call(parent, toLoad, cb)
-  })
+  process.nextTick(loadPlugin.bind(parent), toLoad, cb)
 }
 
 // loads a plugin
--- a/plugin.js
+++ b/plugin.js
@@ -40,6 +40,13 @@ function promise () {
   return obj
 }
 
+function loadPluginNextTick (toLoad, cb) {
+  const parent = this
+  process.nextTick(() => {
+    loadPlugin.call(parent, toLoad, cb)
+  })
+}
+
 function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.started = false
   this.func = func
@@ -49,7 +56,7 @@ function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.timeout = timeout === undefined ? parent._timeout : timeout
   this.name = getName(func, optsOrFunc)
   this.isAfter = isAfter
-  this.q = fastq(parent, loadPlugin, 1)
+  this.q = fastq(parent, loadPluginNextTick, 1)
   this.q.pause()
   this._error = null
   this.loaded = false
--- a/plugin.js
+++ b/plugin.js
@@ -40,13 +40,6 @@ function promise () {
   return obj
 }
 
-function loadPluginNextTick (toLoad, cb) {
-  const parent = this
-  process.nextTick(() => {
-    loadPlugin.call(parent, toLoad, cb)
-  })
-}
-
 function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.started = false
   this.func = func
@@ -250,6 +243,15 @@ Plugin.prototype.finish = function (err, cb) {
   // we start loading the dependents plugins only once
   // the current level is finished
   this.q.resume()
+}
+
+// delays plugin loading until the next tick to ensure any bound `_after` callbacks have a chance
+// to run prior to executing the next plugin
+function loadPluginNextTick (toLoad, cb) {
+  const parent = this
+  process.nextTick(() => {
+    loadPlugin.call(parent, toLoad, cb)
+  })
 }
 
 // loads a plugin
--- a/test/await-after.test.js
+++ b/test/await-after.test.js
@@ -4,6 +4,8 @@ const { test } = require('tap')
 const boot = require('..')
 const { promisify } = require('util')
 const sleep = promisify(setTimeout)
+const fs = require('fs').promises
+const path = require('path')
 
 test('await after - nested plugins with same tick callbacks', async (t) => {
   const app = {}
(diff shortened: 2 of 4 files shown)


plugin.js
```diff
@@ -49,7 +56,7 @@ function Plugin (parent, func, optsOrFunc, isAfter, timeout) {
   this.timeout = timeout === undefined ? parent._timeout : timeout
   this.name = getName(func, optsOrFunc)
   this.isAfter = isAfter
-  this.q = fastq(parent, loadPlugin, 1)
+  this.q = fastq(parent, loadPluginNextTick, 1)
```
chrskrchr, line 59: I realize that delaying the loading until the next tick somewhat defeats the purpose of using a "fast" queue here, but this was the least invasive way I could find to keep things loading in the right order while also fixing the bug described in [fastify/3726](https://github.com/fastify/fastify/issues/3726).

test/await-after.test.js
```diff
@@ -306,6 +308,77 @@ test('await after complex scenario', async (t) => {
...
+
+  async function third () {
+    await sleep(10)
+    thirdLoaded = true
+  }
+
+  function fourth (server, opts, done) {
+    fourthLoaded = true
+    done()
+  }
```
chrskrchr, line 379: This test is nearly identical to the test added in [#173](https://github.com/fastify/avvio/pull/173/files#diff-be4d16e569facb97657f262fcd3701d797567aef795a4a3dcd29b95b08486fe4R285) but with the addition of a fourth synchronous plugin registration function to test the case where the function completes on the same tick as when it was invoked.

plugin.js
chrskrchr: Also, I assumed that incurring a few extra ticks in the boot sequence wouldn't be a huge deal since it's a "one and done" kind of deal, i.e., not on the critical path for request execution.  (but please correct me if that's a bad assumption)

plugin.js
```diff
@@ -40,6 +40,13 @@ function promise () {
   return obj
 }
 
+function loadPluginNextTick (toLoad, cb) {
```
RafaelGSS, line 43: Can you add a comment explaining why this is necessary?

chrskrchr: update

plugin.js
jsumners: > Also, I assumed that incurring a few extra ticks in the boot sequence wouldn't be a huge deal since it's a "one and done" kind of deal, i.e., not on the critical path for request execution. (but please correct me if that's a bad assumption)

I agree, but others are very concerned about boot speed. Are we able to quantify the difference? Something along the lines of https://github.com/jsumners/fjs-ajv-experiment should suffice.

plugin.js
```diff
@@ -245,6 +245,15 @@ Plugin.prototype.finish = function (err, cb) {
   this.q.resume()
 }
 
+// delays plugin loading until the next tick to ensure any bound `_after` callbacks have a chance
+// to run prior to executing the next plugin
+function loadPluginNextTick (toLoad, cb) {
+  const parent = this
+  process.nextTick(() => {
+    loadPlugin.call(parent, toLoad, cb)
+  })
```
mcollina, line 254: please avoid creating this function.

plugin.js
chrskrchr: > Are we able to quantify the difference? Something along the lines of https://github.com/jsumners/fjs-ajv-experiment should suffice.

sure - I'll take a look at adapting your repo to quantify the perf difference

chrskrchr: I still need to convince myself that [my benchmark's plugin registrations](https://github.com/chrskrchr/fastify-avvio-benchmark/blob/main/index.js#L9-L18) are representative of what happens in the real world (so please take these with a grain of salt for now), but my [benchmark script](https://github.com/chrskrchr/fastify-avvio-benchmark) suggests that the overhead added with this change is minimal - less than `0.5%`.

#### Run #1 (+0.47%)
```
┌──────────┬─────────┬──────────┬──────────┬─────────────┬────────────┐
│ (index)  │ samples │ min (ms) │ max (ms) │ median (ms) │ total (ms) │
├──────────┼─────────┼──────────┼──────────┼─────────────┼────────────┤
│ current  │   500   │    33    │    82    │     40      │   20124    │
│ proposed │   500   │    35    │    70    │     40      │   20218    │
└──────────┴─────────┴──────────┴──────────┴─────────────┴────────────┘
```

#### Run #2 (+0.42%)
```
┌──────────┬─────────┬──────────┬──────────┬─────────────┬────────────┐
│ (index)  │ samples │ min (ms) │ max (ms) │ median (ms) │ total (ms) │
├──────────┼─────────┼──────────┼──────────┼─────────────┼────────────┤
│ current  │   500   │    32    │    77    │     37      │   18612    │
│ proposed │   500   │    31    │    56    │     37      │   18691    │
└──────────┴─────────┴──────────┴──────────┴─────────────┴────────────┘
```

#### Run #3 (+0.27%)
```
┌──────────┬─────────┬──────────┬──────────┬─────────────┬────────────┐
│ (index)  │ samples │ min (ms) │ max (ms) │ median (ms) │ total (ms) │
├──────────┼─────────┼──────────┼──────────┼─────────────┼────────────┤
│ current  │   500   │    33    │    83    │     39      │   19473    │
│ proposed │   500   │    33    │    62    │     39      │   19525    │
└──────────┴─────────┴──────────┴──────────┴─────────────┴────────────┘
```

plugin.js
chrskrchr: hey @mcollina - apologies if I'm being dense here, but is the issue here that this work is being delayed until the next tick?  or that it's being done within an anonymous function

plugin.js
jsumners: Looks good to me.

plugin.js
mcollina: Just with the anonymous function. Instead of allocating a new closure you can use a top-level function and pass arguments to it:

```js
process.nextTick(console.log, 42)
```

test/await-after.test.js
Eomm: Could we add a new test case that verifies the user error that caused the rollback?

test/await-after.test.js
```diff
@@ -306,6 +308,77 @@ test('await after complex scenario', async (t) => {
...
+  let fourthLoaded = false
+
+  app.use(first)
+  await app.after()
+  t.ok(firstLoaded, 'first is loaded')
+  t.notOk(secondLoaded, 'second is not loaded')
+  t.notOk(thirdLoaded, 'third is not loaded')
+  t.notOk(fourthLoaded, 'fourth is not loaded')
+
+  const contents = await fs.readFile(path.join(__dirname, 'fixtures', 'dummy.txt'), 'utf-8')
```
RafaelGSS, line 328: what is your intention reading it twice?

plugin.js
chrskrchr: oh wow - I had no idea that `process.nextTick()` could accept a list of args 🤦 

I just push a commit with this refactor.  The function still needs to use `.bind(parent)` to set the right context for the `loadPlugin` call (which still creates a new function internally), but the refactor is a bit more readable IMO.

test/await-after.test.js
chrskrchr: Just wanting to mix in a bit of async FS I/O in various contexts (e.g., both at the top-level and within a plugin) per @Eomm's request on the original PR: 

https://github.com/fastify/avvio/pull/173#discussion_r814227953

RafaelGSS (changes requested):
The tests added are passing even without your change. Could you take a look?

chrskrchr:
> The tests added are passing even without your change. Could you take a look?

ooof - that's obviously unexpected.  Will take a look ASAP.

test/await-after.test.js
chrskrchr: OK - so it looks like this top-level async FS read (a late addition to the original PR) was what caused the test to pass without my changes.  I've removed the top-level read and verified that test fails as expected without my changes.

@Eomm - can you verify that the remaining FS read that occurs within the plugin registration function on L363 still satisfies [your original request to mix in some async FS I/O in the test](https://github.com/fastify/avvio/pull/173#discussion_r814227953)?

xtx1130 (changes requested):
I think this case can be added in test cases: 
```js
const t = require('tap')
const avvio = require('avvio')
let record = ''
const app = {}
avvio(app, { autostart: false })
function sleep(timeMs){
  return new Promise((resolve) => setTimeout(resolve, timeMs));
}

(async () => {
  await app.use(async function plugin0() {
    t.pass("plugin0 init");
    record += 'plugin0|'
  })
  await app.use(async function plugin1() {
    t.pass("plugin1 init");
    await sleep(500);
    record += 'plugin1|'
  })
  await sleep(1);
  await app.use(async function plugin2() {
    t.pass("plugin2 init");
    await sleep(500);
    record += 'plugin2|'
  })
  record += 'ready'
  t.equal(record, 'plugin0|plugin1|plugin2|ready')
})()
```
before:
image

after:
image

chrskrchr:
> It's more or less testing the same behavior AFAICT, but with a different style of assertions.

Actually, I see that your test case is actually testing the `await app.use()` behavior rather than `app.after()`.  I've added it to the `await-use.test.js` suite.

chrskrchr:
> > The tests added are passing even without your change. Could you take a look?
> 
> ooof - that's obviously unexpected. Will take a look ASAP.

Also, the false negative mentioned above was addressed by removing the top-level async FS read that was preventing the underlying bug from occurring when my code wasn't present:

https://github.com/fastify/avvio/pull/177#discussion_r817729438

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

chrskrchr:
> Merged #177 into master.

😬 I mean, what are the odds I screw this up _twice_?

xtx1130 (commented):
Great work!

PR #189, opened 2022-07-11
Make sure we always start in a microtick

mcollina:
Signed-off-by: Matteo Collina <hello@matteocollina.com>
Fixes: https://github.com/fastify/fastify/issues/4125

#### Checklist

### DIFF
--- a/test/await-use.test.js
+++ b/test/await-use.test.js
@@ -270,3 +270,25 @@ test('await use - mix of same and future tick callbacks', async (t) => {
   record += 'ready'
   t.equal(record, 'plugin0|plugin1|plugin2|ready')
 })
+
+test('await use - fork the promise chain', { only: true }, (t) => {
+  t.plan(3)
+  const app = {}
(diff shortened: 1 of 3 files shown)


zekth (approved):
lgtm

test/await-use.test.js
```diff
@@ -270,3 +270,25 @@ test('await use - mix of same and future tick callbacks', async (t) => {
   record += 'ready'
   t.equal(record, 'plugin0|plugin1|plugin2|ready')
 })
+
+test('await use - fork the promise chain', { only: true }, (t) => {
```
stefee, line 274: What does only:true mean?

delvedor: It should be removed, it tells to tap to **only** run this test.

mcollina: Nothing. If it's run with TAP_ONLY=true env variable, it would run only this test. I'll remove, good spot!

PR #199, opened 2023-01-09
feat: rewrite

### LINKED ISSUE #118: Rewrite avvio from scratch
Avvio is a fantastic library and the hearth of the Fastify plugin system. In the first days of Fastify (4 years ago!!) @mcollina and I spent months working on this library, trying to figure out all the moving pieces, and I'm very happy about the outcome.
Avvio is predictable, works well with good enough performances, and has a comprehensive set of features.
Unfortunately, in the past two years, we have focused more on adding features and less on refactoring the library. I guess we can all agree it is something that should be done in the near future, as the internals has become rather complex both to understand and maintain.

This is a long term goal, but I would love to see a complete rewrite of the library by Fastify v4.
The first thing we should do, in my opinion, is to define which features are essential and which are no longer useful or too hard to maintain.
For example, `register` (`use` here) being awaitable but also returning Fastify itself is nice but complex to maintain and with its fair amount of edge cases (see https://github.com/fastify/avvio/pull/107, https://github.com/fastify/avvio/pull/114, and https://github.com/fastify/avvio/pull/116). Or `.after`, which was very useful in the past, but today it's just a difficult component to understand for users with its edge cases as well(see https://github.com/fastify/fastify-express/issues/3).

Once we have defined which features we should keep, we should write a Fastify application that uses them all, so when we'll start the refactoring we'll make sure that all of them will still work.
We'll use this issue for tracking the effort, let's keep here the discussion about which feature should be kept and then open a separate pr for each one.
It would be nice also to ask the community how are they using the plugin system, so we are sure not to break the most used parts.

cc @fastify/core

climba03003:
Resolves #118 

#### TODO
- [x] group the code that provide same feature as function
- [x] split into multiple files for better management
- [x] variables name to be consistence over the package and use full name
- [ ] add comments to explain the details
- [x] remove `.express`
- [x] remove `.after`
- [ ] split `.ready` and `.start`

#### Checklist


mcollina:
Thanks for taking this on!

lilpoozie2005 (approved):

climba03003:
Using `Promise.try` to simplify the codebase.
Removed `.after` and replace with `.then`.

Pending to add the `toJSON` and `prettyPrint` back.

Tony133:
@mcollina @climba03003 Can we consider this a PR for Fastify v6 ? It seems to me that the code has been cleaned up and simplified.

climba03003:
> Can we consider this a PR for Fastify v6 ?

It can be closed, the rewrite should be started based on the current main.

PR #214, opened 2023-06-23
remove then-sleep dev dependency

Uzlopak:
Why use a package if we can use native promises and it is basically a three liner.

#### Checklist


Fdawgs (approved):

voxpelli:
And from Node.js v15 this is even built into Node.js: https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options

PR #219, opened 2023-06-24
use @fastify/error

Uzlopak:
alternative to #197 

@mcollina asked if we can use @fastify/error directly. This PR uses @fastify/error instead of AVVIO own implementation. Only difference is, that the name of the Errors will be FastifyError instead of AvvioError.

#### Checklist


mcollina (approved):
lgtm

PR #218, opened 2023-06-24
extract create-promise

Uzlopak:
based on #199 

added unit tests and jsdoc

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #221, opened 2023-06-25
use named export for timeTree, use null as initialization value for TimeTreeNode.stop

Uzlopak:
Based on #218 I assume that we should use null as initializiation value for stop. 
Also making named export for TimeTree to keep the style we use in the new files ;)

#### Checklist


mcollina (approved):
lgtm

PR #225, opened 2023-06-27
Refactor plugin

Uzlopak:
Refactoring Plugin.

Trying to use more meaningful parameter names. And also kind of decoupling avvio/boot from plugin. 

I decoupled plugin from boot/avvio and added a new Error AVV_ERR_PLUGIN_EXEC_TIMEOUT

loadPlugin is now a method of boot/avvio.

#### Checklist


lib/load-plugin.js
```diff
@@ -31,12 +32,32 @@ function loadPlugin (instance, plugin, callback) {
   // place the plugin at the top of _current
   instance._current.unshift(plugin)
 
-  plugin.exec((last && last.server) || instance._server, (err) => {
+  if (instance._error && !plugin.isAfter) {
```
Uzlopak, line 35: I moved this from plugin.exec, as it seemed odd to me, that we could do this already in the loadPlugin function, and decouple so Plugin from Avvio.

lib/plugin.js
```diff
@@ -82,55 +106,59 @@ Plugin.prototype.exec = function (server, cb) {
...
 
   if (this.timeout > 0) {
     debug('setting up timeout', name, this.timeout)
     timer = setTimeout(function () {
       debug('timed out', name)
       timer = null
-      const err = new AVV_ERR_READY_TIMEOUT(name)
-      err.fn = func
-      done(err)
+      const readyTimeoutErr = new AVV_ERR_READY_TIMEOUT(name)
```
Uzlopak, line 117: Wouldnt it make more sense to create and use a AVV_ERR_PLUGIN_EXEC_TIMEOUT?

lib/plugin.js
```diff
@@ -82,55 +106,59 @@ Plugin.prototype.exec = function (server, cb) {
...
+      readyTimeoutErr.fn = func
```
Uzlopak, line 118: I think the only reason we store func as fn to the instance of the timeoutErr, is that we check in plugin-timeout.test.js if the right function was the reason of the error. 
It has imho no real benefit. 

Maybe we should just remove this fn assignment.

lib/plugin.js
```diff
@@ -1,76 +1,100 @@
...
+  this.started = false
+  /**
+   * @type {string}
+   */
+  this.name = getPluginName(func, options)
+
+  /**
+   * @type {fastq.queue}
+   */
+  this.queue = fastq(parent, loadPluginNextTick, 1)
```
Uzlopak, line 45: the only reason, we have to pass parent to the Plugin-constructor is that we have to instantiate a fastq-instance. 

Any idea how we could decouple it completely?

Uzlopak:
@mcollina 
@Eomm 
@climba03003 

I have some question and ask for feedback ;)

Uzlopak:
@metcoder95 
@jsumners 

Do you have maybe some input regarding my questions?

Uzlopak:
I decoupled plugin from boot/avvio and added a new Error AVV_ERR_READY_TIMEOUT

lib/plugin.js
mcollina: Add a TODO and possibly do it in a separate PR. In case we break something it's easier to revert.

lib/plugin.js
mcollina: I don't think it's possible to decouple them

lib/plugin.js
mcollina: Why?

Uzlopak: Imho there are two timeouts. One is that the plugin loaded too long. For this I throw AVV_ERR_PLUGIN_EXEC_TIMEOUT. The other timeout is when the overall loading takes simply too long, e.g. all plugins take together more time than the avvio is allowed to.

I think it should be differenciated. But I am willing to revert that change.

lib/plugin.js
Uzlopak: I think i successfully decoupled them. Now I pass a fastq queue as first param instead of the avvio instance.

lib/errors.js
```diff
@@ -26,5 +26,9 @@ module.exports = {
   AVV_ERR_READY_TIMEOUT: createError(
     'AVV_ERR_READY_TIMEOUT',
     "Plugin did not start in time: '%s'. You may have forgotten to call 'done' function or to resolve a Promise"
+  ),
+  AVV_ERR_PLUGIN_EXEC_TIMEOUT: createError(
+    'AVV_ERR_PLUGIN_EXEC_TIMEOUT',
```
metcoder95, line 31: The change is sound to me but they are really similar by now. Shall we change the message to differentiate?

metcoder95 (commented):
Seeing the changes, I'm starting to think if it would be worth it to explore a further change to the data structure that manages the plugins. Currently is a `queue` (that slightly mimics a `linked list` in the sense that they are more or less aware of each other).

Instead, if we can successfully shift it to a `tree`, we would be able to enable several functionalities (like parallel loading) by identifying plugin dependants to locate nodes that can be loaded in parallel (does not depend on other plugins, a `leaf` node basically).

Just thinking out loud, but is it something we can explore further in another PR?

jsumners:
> @metcoder95 @jsumners
> 
> Do you have maybe some input regarding my questions?

I do not have the mental capacity right now to study this codebase and provide useful feedback.

boot.js
```diff
@@ -367,6 +370,71 @@ Boot.prototype.toJSON = function () {
...
+ * @param {Error} [err]
+ */
+
+/**
+ * Load a plugin
+ *
+ * @param {Plugin} plugin
+ * @param {LoadPluginCallback} callback
+ */
+Boot.prototype.loadPlugin = function (plugin, callback) {
```
mcollina, line 384: I think this should be `_loadPlugin`?

boot.js
```diff
@@ -367,6 +370,71 @@ Boot.prototype.toJSON = function () {
...
+      callback(err)
+    })
+  }
+}
+
+/**
+* Delays plugin loading until the next tick to ensure any bound `_after` callbacks have a chance
+* to run prior to executing the next plugin
+*/
+Boot.prototype.loadPluginNextTick = function (plugin, callback) {
```
mcollina, line 434: I think this should be `_loadPluginNextTick`?

Uzlopak: done ;)

lib/plugin.js
Uzlopak: @metcoder95 

We should discuss here if the new Error makes more sense.

lib/errors.js
Uzlopak: see comment above ;)

lib/plugin.js
Uzlopak: done

Uzlopak:
@jsumners 
Thanks anyway :)

@metcoder95
This could be possible. But I am currently thinking about how I could increase the code coverage of plugin.js as it is still at poor 87%. ;D

mcollina (approved):
lgtm

metcoder95 (approved):
LGTM 👍

Uzlopak:
Thank you. I will hopefully provide more PRs today. ;)

PR #235, opened 2023-08-14
chore: move wrap to be method _expose()

Uzlopak:
Make wrap() function a method `_expose()`

#### Checklist


mcollina (approved):
lgtm

mcollina:
Possibly it might be better to move this into its separate file and keep it a function. Boot has a already too many responsibilities.

Uzlopak:
I plan to reduce the complexity of boot. Here the original wrap function makes much more sense, to be part of boot class, as it is 99% about boot and the modification of the underlying server with methods of boot.

PR #253, opened 2024-04-29
workflow: pin node versions to LTS

gurgunday:


gurgunday:
@mcollina do you know why this could be?

mcollina:
windows 22 should be skipped for now

PR #278, opened 2025-01-02
test: migrated catch-override-exception.test.js from tap to node:test

liuhanqu:


test/catch-override-exception.test.js
```diff
@@ -1,26 +1,26 @@
...
+  t.plan(1)
 
   const server = {}
 
   const app = boot(server, {
     autostart: false
   })
   app.override = function () {
-    throw Error('catch it')
+    throw new Error('catch it')
```
climba03003, line 15:
```suggestion
    throw Error('catch it')
```

liuhanqu: throw an error without `new` keyword?

climba03003: There is no different with or without the `new`.
So, the change itself is unnecessary.

climba03003 (approved):

test/catch-override-exception.test.js
```diff
@@ -1,10 +1,10 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const boot = require('..')
 
-test('catch exceptions in parent.override', (t) => {
-  t.plan(2)
+test('catch exceptions in parent.override', (t, testCompleted) => {
```
Uzlopak, line 6: Should be called done

climba03003: It is used to prevent mixing up with the avvio `done` callback.
https://github.com/fastify/avvio/pull/277#discussion_r1898357599

liuhanqu: as suggest by @Eomm  in https://github.com/fastify/avvio/pull/277, i think we should keep the name of second variable consistent

Uzlopak: @Eomm had an issue with the fact it was called `done_` I guess, it should have been called `testDone`.

liuhanqu: It's ok to use `testDone` or `testCompleted`,  we should keep consistent in all tests.
If use `testDone`,  i would replace `testCompleted` with `testDone` in `basic.test.js`  and `callbacks.test.js`

test/catch-override-exception.test.js
```diff
@@ -20,7 +20,7 @@ test('catch exceptions in parent.override', (t) => {
     .start()
 
   app.ready(function (err) {
-    t.type(err, Error)
-    t.match(err, /catch it/)
+    t.assert.strictEqual(err.message, 'catch it')
+    testDone()
```
gurgunday, line 24: Can we do this instead of `t.type`?

```suggestion
    t.assert.ok(err instanceof Error)
    t.assert.strictEqual(err.message, 'catch it')
    testDone()
```

test/catch-override-exception.test.js
```diff
@@ -1,10 +1,10 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const boot = require('..')
 
-test('catch exceptions in parent.override', (t) => {
-  t.plan(2)
+test('catch exceptions in parent.override', (t, testDone) => {
+  t.plan(1)
```
Eomm, line 7:
```suggestion
  t.plan(2)
```

Eomm (approved):

PR #279, opened 2025-01-02
test: migrated after-and-ready.test.js from tap to node:test

liuhanqu:


climba03003 (approved):

test/after-and-ready.test.js
```diff
@@ -1,60 +1,62 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const boot = require('..')
 
-test('boot a plugin and then execute a call after that', (t) => {
+test('boot a plugin and then execute a call after that', (t, testCompleted) => {
```
Uzlopak, line 6: should be called done, like it is named in the docs for node native test runner

https://nodejs.org/api/test.html#test-runner

liuhanqu: I have used `testDone` instead of `testCompleted`

gurgunday (approved):
lgtm

PR #312, opened 2025-06-18
test: migrated pretty-print.test.js to tap to node:test

Tony133:
#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

Proposal:

- migrated `pretty-print.test.js` from `tap` to `node:test` 🔥

### DIFF
--- a/test/pretty-print.test.js
+++ b/test/pretty-print.test.js
@@ -3,7 +3,7 @@
 const { test } = require('node:test')
 const boot = require('..')
 
-test('pretty print', async t => {
+test('pretty print', (t, done) => {
   t.plan(19)
 
   const app = boot()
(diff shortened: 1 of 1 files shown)


test/pretty-print.test.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const boot = require('..')
 
-test('pretty print', t => {
+test('pretty print', async t => {
```
jean-michelet, line 6:
```suggestion
test('pretty print', (t, done) => {
```

Can we keep using the callback api and call `done` inside?

jean-michelet (commented):
Thanks for the PR.

We'll see what other collaborators think, but I personally see no interest to promisify the callback.

test/pretty-print.test.js
Tony133: > Can we keep using the callback api and call done inside?

yes, we can also use callback api, I did a commit here: [4fc9916](https://github.com/fastify/avvio/pull/312/commits/4fc99166babb0c52412da840e209dc57dc5d37cf)


the first commit is with Promises the second is with callbacks, so there are two ways, to do the test 😉

jean-michelet (approved):

mcollina (approved):
lgtm

PR #324, opened 2025-08-16
chore: remove tap from tests

ilteoood:
#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/test/after-throw.test.js
+++ b/test/after-throw.test.js
@@ -1,24 +1,39 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const boot = require('..')
+const { setTimeout } = require('timers/promises')
 
-test('catched error by Promise.reject', (t) => {
+test('catched error by Promise.reject', async (t) => {
   const app = boot()
   t.plan(2)
 
-  t.threw = function (err) {
-    t.equal(err.message, 'kaboom2')
-  }
+  const uncaughtExceptionHandlers = process.listeners('uncaughtException')
 
-  app.use(function (f, opts) {
-    return Promise.reject(new Error('kaboom'))
-  }).after(function (err) {
-    t.equal(err.message, 'kaboom')
-    throw new Error('kaboom2')
-  })
+  process
+    .removeAllListeners('uncaughtException')
+    .once('uncaughtException', (err) => {
+      t.assert.strictEqual(err.message, 'kaboom2')
 
+      // Restore original handlers
+      uncaughtExceptionHandlers.forEach((handler) =>
+        process.on('uncaughtException', handler)
+      )
+    })
+
+  app
+    .use(function (f, opts) {
+      return Promise.reject(new Error('kaboom'))
+    })
+    .after(function (err) {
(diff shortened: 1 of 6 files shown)


test/to-json.test.js
```diff
@@ -1,11 +1,41 @@
...
+    return
+  }
+
+  if (isExpectedRegex) {
+    t.assert.match(actual.toString(), expected)
+    return
+  }
+
+  t.assert.strictEqual(actual, expected)
+}
```
Uzlopak, line 35: what is this?

test/after-throw.test.js
```diff
@@ -1,24 +1,39 @@
...
 
+  app
+    .use(function (f, opts) {
+      return Promise.reject(new Error('kaboom'))
+    })
+    .after(function (err) {
+      t.assert.strictEqual(err.message, 'kaboom')
+      throw new Error('kaboom2')
+    })
+
```
Uzlopak, line 32: what is this deviation?

Uzlopak (changes requested):
Seem to be wrong. Will check now in my local development.

test/to-json.test.js
ilteoood: node test runner doesn't implement match in the same way tap does.
Too keep the `outJson` structure, I've created this companion function that mimics the behaviour of tap's match.

test/after-throw.test.js
ilteoood: for tap the test was overriding the `threw` function, but this is not possible for node test runner.
The only idea that came into my mind is to add a single `uncaughtException` that does the check we need.

Uzlopak: This is SUPER strange behavior. I would not expect an uncaught promise rejection. It sounds more like a bug whch was covered by tap.

ilteoood:
Thank you so much for your help @Uzlopak.
I have inserted also a final change to remove tap from thenify. Please let me know your thoughs about it.

Uzlopak:
@ilteoood 

Also. Thank you for your work.

@mcollina 
@Fdawgs 
@Eomm 
@climba03003 
@jsumners 

it seems that there was a unhandled promise rejection swallowed by the tests. I modified the function to catch it. Can you please have a look? Its a quite critical part of the code i think.

mcollina (approved):
lgtm

Uzlopak (approved):

test/esm.mjs
```diff
@@ -1,12 +1,12 @@
-import { test } from 'tap'
+import { test } from 'node:test'
 import boot from '../boot.js'
 
-test('support import', async (t) => {
+await test('support import', async (t) => {
```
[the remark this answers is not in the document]
Uzlopak: I tested it locally and it does complete?

I also added a t.plan(1) just to be sure.

PR #351, opened 2026-07-14
fix: avoid stack overflow in ready callbacks

### LINKED ISSUE #6843: ready() hangs indefinitely with 400+ routes with a plugin that uses onReady, like swagger
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.10.0

### Plugin version

9.8.0

### Node.js version

24.15.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

13

### Description

Our app stops coming up if you add an extra route: it reaches `await app.ready()`, which never resolves.

The minimal cause seems to be:
1. we register the fastify swagger plugin
2. we add hundreds of routes

"hundreds" means about 400 with two plugins, or about 800 with one plugin, on the node 24 defaults on my machine.

My understanding of the problem is that:
 * the internals of `avvio` / `fastq` are recursive (i.e. require increasing amounts of stack for increasing numbers of registrations), and 
 * poorly handle a stack overflow being thrown, leading to a 
 * corrupt internal state, and 
 * the queue never being marked as empty, so
 * the promise never resolves.

How and where this happens varies based on stack size, see the readme in the repo.

I've raised this against the fastify repo because I think a good fix could be in https://github.com/fastify/fastify/blob/94bcbcc6e2ef3b8e8f8e8797fe551ccbe7b942fd/lib/hooks.js#L159-L166 (i.e. make it crash harder on internal node errors or syntax errors).

### Link to code that reproduces the bug

https://github.com/FauxFaux/fastify-registration-hang

### Expected Behavior

`app.ready()` shouldn't hang indefinitely.

Ideally, we shouldn't need to change the stack size to register more routes.

Alternatively, the stack overflow should propagate out of `app.ready()`, or it should give some other meaningful failure.

mcollina:
## Summary

- defer callback-style ready queue advancement to the next tick to avoid recursive fastq execution
- guard ready callback completion so duplicate calls cannot corrupt queue state
- add regression coverage with and without ready timeouts

## Background

Fastify's callback-style `onReady` wrapper catches exceptions thrown by `done()`. Because Avvio advanced the ready queue synchronously, enough callbacks could overflow the stack inside downstream queue processing. Fastify then called `done(error)` a second time, potentially leaving fastq in a state where `ready()` never settled.

Deferring queue advancement makes the ready queue stack-safe and ensures Fastify's `try/catch` only observes errors from the hook itself.

Fixes fastify/fastify#6843

## Tests

- `npm test`
- `npm run lint`
- Fastify Swagger reproduction with 400 and 1,000 routes under `node --stack-size=256`


Tony133 (approved):

climba03003 (approved):
Double checked with the original repro.
It does fix the issue.

LGTM.

